PR #1, opened 2017-08-23
Deploy scripts

gunzip:
- Added script to deploy functions to the Azure cloud
- Added docs (README.md)


README.md
```diff
@@ -5,3 +5,116 @@
...
+
+There a few options to deploy the Functions to Azure.
+
+### Using GitHub
+
+If you have a GitHub account you can set up a [continuos deployment flow](https://docs.microsoft.com/en-us/azure/azure-functions/functions-continuous-deployment).
+
+1. Fork the GitHub repository into your account.
+
+2. Edit scripts/local.config.js and write up your GitHub settings:
```
cloudify: use backticks for file paths

package.json
```diff
@@ -28,6 +28,7 @@
     "express": "^4.15.3",
     "nodemailer": "^4.0.1",
     "nodemailer-sendgrid-transport": "^0.2.0",
+    "shelljs": "^0.7.8",
```
cloudify: move this to the `devDependencies` block

scripts/resources.js
```diff
@@ -0,0 +1,297 @@
+const $ = require('shelljs')
```
cloudify: no shebang? is this script meant to be executed?

cloudify: also the script needs a comment block explaining it's purpose and usage

scripts/resources.js
```diff
@@ -0,0 +1,297 @@
...
+const program = require('commander')
+const CONF = require('./local.config')()
+
+// die on first error
+$.config.fatal = true
+
+// VERBOSE=1 node deploy.js
+$.config.verbose = !!process.env.VERBOSE || false
```
cloudify: there are two problems with this statement :)

`!!"0" === true`

also the `|| false` is unnecessary (`A || false === A`)

scripts/resources.js
```diff
@@ -0,0 +1,297 @@
...
+
+// Create a resource group if not exists
+// az group create \
+//       --name $resourceGroupName \
+//       --location $location
+
```
cloudify: a sanity check would be nice:

```
if (!shell.which('az')) {
  shell.echo('Sorry, this script requires the Azure management cli to be installed');
  shell.exit(1);
}
```

scripts/resources.js
cloudify: also `$` is not a very nice name for a variable, perhaps `shell` ?

scripts/resources.js
```diff
@@ -0,0 +1,297 @@
...
+    --resource-group ${CONF.resourceGroupName} \
+    --slot-settings "SENDGRID_KEY=${CONF.sendGridKey}"`)
+}
+
+//////////////////// FUNCTIONS (git deployment)
+
+// az functionapp deployment user set --user-name $username --password $password
+
+const deploy_functionDeployementSource = () => {
+  if (CONF.gitRepoUrl !== '') {
```
cloudify: also `CONF.gitRepoUrl != null`

scripts/resources.js
gunzip: about the shebang, it does not work in win32 environments (without cygwin / mingw), so
i suggest to run the script using node from command line (which is cross platform)

scripts/resources.js
gunzip: done.

gunzip:
see https://github.com/teamdigitale/digital-citizenship/pull/14

PR #3, opened 2017-08-24
Adds update with versioning to Profile Model

cloudify:


lib/models/profile.ts
```diff
@@ -51,24 +70,56 @@ export class ProfileModel {
...
-  public createOrUpdateProfile(profile: INewProfile): Promise<IRetrievedProfile | null> {
+  public createProfile(profile: IProfile): Promise<IRetrievedProfile> {
+    return new Promise((resolve, reject) => {
+      // the first version of a profile is 0
+      const initialVersion = 0;
+      // the ID of each profile version is composed of the profile ID and its version
+      // this makes it possible to detect conflicting updates (concurrent creation of
+      // profiles with the same profile ID and version)
+      const profileId = generateVersionedProfileId(profile.fiscalCode, initialVersion);
+      DocumentDbUtils.createDocument(
```
gunzip: maybe we can just `return  DocumentDbUtils.createDocument()` without promise wrapping

lib/utils/error_handler.ts
```diff
@@ -5,10 +5,29 @@ import * as express from "express";
  *
  * @param response The Express Response for the current request
  */
-export function handleErrorAndRespond(response: express.Response): (reason: any) => any {
-  return (reason: any) => {
-    response.status(500).json({
+export function handleErrorAndRespond<T>(response: express.Response): (reason: T) => express.Response {
```
gunzip: probably it's better to have an [express error handler](http://expressjs.com/en/guide/error-handling.htm) so you don't have to call it explicitly, you just `throw new ApiError(status, reason)` inside the controller.

lib/utils/error_handler.ts
```diff
@@ -5,10 +5,29 @@ import * as express from "express";
...
     });
   };
 }
+
+/**
+ * Returns an error handler that responds to the current HTTP request.
+ *
+ * @param response The Express Response for the current request
+ */
+export function handleNullableResultAndRespond<T, E>(
```
gunzip: probably it's better to do this with an express middleware, so you don't have to call it from inside the controller function (you can pass the middleware to the routing function).

PR #5, opened 2017-08-25
Remove forked packages (azure-function-types and azure-functions-express)

gunzip:
- there 's no need to fork azure-function-express as we can pass context to the app and retrieve it from controllers
- fixed IContext type with logging facilities
- get rid of azure-function-type fork in workdir, just use the version from npm
- now you can use console.log when running locally and it will call context.log ;)

All test passes.

We have to converge on same style for newlines (using [prettier](https://github.com/prettier/prettier-vscode) maybe ?)


gunzip:
I've merged this with the no-any branch and backported the context<binding> part

PR #29, opened 2017-09-07
150777355 test organization model

lussoluca:


lib/models/__tests__/organization.test.ts
```diff
@@ -0,0 +1,202 @@
...
+        const result = await model.findByOrganizationId("id");
+
+        expect(result.isRight).toBeTruthy();
+        if (result.isRight) {
+            expect(result.right.isDefined).toBeTruthy();
+            expect(result.right.get).toEqual("result");
+        }
+    });
+
+    it("should resolve a promise to undefined if no organization is found", async () => {
```
cloudify: Actually we need to change this description (also in the other tests) since I've opted for `Options` instead of `undefined` values.

It should say `should resolve a promise to an empty value if no organization is found`

lib/models/__tests__/organization.test.ts
```diff
@@ -0,0 +1,202 @@
...
+        expect(clientMock.createDocument.mock.calls[0][2]).toHaveProperty("partitionKey", organizationId);
+        expect(result.isRight).toBeTruthy();
+        if (result.isRight) {
+            expect(result.right.organizationId).toEqual(newOrganization.organizationId);
+            expect(result.right.id).toEqual(`${organizationId}-${"0".repeat(16)}`);
+            expect(result.right.version).toEqual(0);
+        }
+    });
+
+    it("should reject the promise in case of error", async () => {
```
cloudify: also here the promise is actually resolved to an error value (a `Left<DocumentDb.QueryError>`)

lib/models/__tests__/organization.test.ts
```diff
@@ -0,0 +1,202 @@
...
+            expect(result.right.isDefined).toBeTruthy();
+            const updatedOrganization = result.right.get;
+            expect(updatedOrganization.organizationId).toEqual(aRetrievedOrganization.organizationId);
+            expect(updatedOrganization.id).toEqual(`${organizationId}-${"0".repeat(15)}1`);
+            expect(updatedOrganization.version).toEqual(1);
+            expect(updatedOrganization.name).toEqual("MyOrganization");
+        }
+    });
+
+    it("should reject the promise in case of error (read)", async () => {
```
cloudify: also here, resolved to an error

lib/models/__tests__/organization.test.ts
```diff
@@ -0,0 +1,202 @@
...
+        expect(clientMock.readDocument).toHaveBeenCalledTimes(1);
+        expect(clientMock.createDocument).not.toHaveBeenCalled();
+
+        expect(result.isLeft).toBeTruthy();
+        if (result.isLeft) {
+            expect(result.left).toEqual("error");
+        }
+    });
+
+    it("should reject the promise in case of error (create)", async () => {
```
cloudify: ditto

lib/models/organization.ts
```diff
@@ -39,6 +39,10 @@ function toRetrieved(result: DocumentDb.RetrievedDocument): IRetrievedOrganizati
   } as IRetrievedOrganization);
 }
 
+function getModelId(o: IOrganization): ModelId {
```
cloudify: thanks!

PR #34, opened 2017-09-11
Avoid forks of azure npm packages

gunzip:
We can just use "azure-function-express" and "azure-functions-types" from npm registry.


lib/utils/middlewares/context_middleware.ts
```diff
@@ -1,8 +1,20 @@
...
+import { IContext } from "azure-function-express";
+
+const CONTEXT_IDENTIFIER = "context";
+
+export function setAppContext(app: express.Express, context: IContext<{}>): void {
+  app.set(CONTEXT_IDENTIFIER, context);
+}
+
+export function getAppContext<T>(request: express.Request): IContext<T> {
+  return request.app.get(CONTEXT_IDENTIFIER);
```
cloudify: the return type is actually `IContext<T> | undefined`

gunzip: according to the comment "TODO: validate that the context is indeed defined, respond with ResponseErrorInternal instead"

does it return ResponseErrorInternal  if we throw an Exception ?

because in this case we can just throw, leave the return type to IContext<T> and the check will be satisfied (this is a good spot for throwing)

gunzip: ok I've checked it and it returns internel server error in case context is not found:

wrapRequestHandler|ERROR|http://localhost:7071/api/v1/messages/... |Error: Cannot get context from request

lib/utils/middlewares/context_middleware.ts
```diff
@@ -1,19 +1,32 @@
...
  *
  * @param T The type of the bindings found in the context.
- *
- * TODO: validate that the context is indeed defined, respond with ResponseErrorInternal instead
  */
 export function ContextMiddleware<T>(): IRequestMiddleware<never, IContext<T>> {
-  return (request: IRequestWithContext<T>) => {
-    return Promise.resolve(right(request.context));
+  return (request: express.Request) => {
+    const context: IContext<T> = getAppContext<T>(request)
```
cloudify: better return a left with an `IResponseErrorInternal` here or reject the Promise, middlewares are not supposed to throw exceptions.

gunzip: Anyway middleware cannot prevent throwing an exception in case of *unexpected errors* (ie. in called functions), and probably they should throw in this cases. 

While Eithers are good for *expected* errors, I think that every error that causes an 'internal server error' should be an Exception (application should halt). 

Moreover that's exactly the reason why wrapRequestHandler() handles any Exception returned by middlewares returning a ResponseErrorInternal:

https://github.com/teamdigitale/digital-citizenship-functions/blob/cfdbcd03ad78864d0267a380bda5fd88f4006dbf/lib/utils/request_middleware.ts#L14

cloudify: not really, `wrapRequestHandler` handles rejected Promises, it will not catch the exception :)

also, [exceptions should be avoided](https://www.ibm.com/developerworks/library/j-ft13/index.html) since they hijack the execution flow :)

gunzip: According to the output:

wrapRequestHandler|ERROR|http://localhost:7071/api/v1/messages/... |Error: Cannot get context from request

wrapRequestHandler do catch the Exception.

cloudify: nop, just look at the code, there's no `catch`, the exception is catched by [this Promise](https://github.com/teamdigitale/digital-citizenship-functions/blob/master/lib/utils/request_middleware.ts#L135), that is basically the same as rejecting the Promise.

As a matter of style, it is better to use a functional style when handling exception, the control flow is simpler tu understand (no side effect) and it also retain the type of error (that you instead lose with exceptions and rejected promises).

lib/utils/middlewares/context_middleware.ts
```diff
@@ -1,19 +1,32 @@
...
 
 /**
  * Returns a request middleware that extracts the Azure request context
  * from the request.
  *
  * @param T The type of the bindings found in the context.
- *
- * TODO: validate that the context is indeed defined, respond with ResponseErrorInternal instead
  */
 export function ContextMiddleware<T>(): IRequestMiddleware<never, IContext<T>> {
```
cloudify: for instance, by looking at this function's signature, you cannot know whether this function throws an exception or even what the exception type is, this is also a problem with `Promise<T>` that doesn't have a type for a failed promise, that's the reason I used `Promise<Either<E, T>>`

lib/utils/middlewares/__tests__/context_middleware.test.ts
```diff
@@ -11,21 +11,11 @@ describe("ContextMiddleware", () => {
...
-          test: "test",
-        },
+      app: {
+        get: (s) => ({ context: {} }),
       },
     } as any);
-
     expect(response.isRight).toBeTruthy();
-    if (response.isRight) {
-      expect(response.right).toEqual({
```
cloudify: you should test that `response.right` is equal to the context at row 15, perhaps by returning a non-empty object there and testing for equality.

lib/utils/middlewares/__tests__/context_middleware.test.ts
```diff
@@ -10,12 +10,23 @@ describe("ContextMiddleware", () => {
 
   it("should extract the context from the request", async () => {
     const middleware = ContextMiddleware<ITestBindings>();
-    const response = await middleware({
+
+    const context = {
```
cloudify: one `context` too much? shouldn't it be:

```
const context = {
  log: () => true,
};
```

PR #35, opened 2017-09-11
Refactor toUserGroup

gunzip:
Proposal: refactor toUserGroup function to one-liner (still type safe).

If we need to lookup enum members from string keys (and we do), probably it's better to avoid 'const' enum and just use plain enum that comes with "free" lookup facilities.


cloudify:
Thanks, can you please rebase and squash?

PR #37, opened 2017-09-13
Make it work with applicationinsights 0.22

gunzip:
It looks like TelemetryClient works only for numeric (integer) properties (see https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/Library/TelemetryTypes/EventTelemetry.ts)

Probably it's safer to switch back to "defaultClient" (without creating new client instances with 'new')

### DIFF
--- a/lib/controllers/messages.ts
+++ b/lib/controllers/messages.ts
@@ -212,11 +212,14 @@ export function CreateMessageHandler(
     if (messagePayload.dry_run) {
       // if the user requested a dry run, we respond with the attributes
       // that we received
-      applicationInsightsClient.trackEvent({name: eventName, properties: {
+      applicationInsightsClient.trackEvent({
+        name: eventName,
+        properties: {
           ...eventData,
           dryRun: "true",
           success: "true",
-      }});
+        },
+      });
       const response: IResponseDryRun = {
         bodyShort: messagePayload.body_short,
         senderOrganizationId: userOrganization.organizationId,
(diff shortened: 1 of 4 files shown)


lib/controllers/messages.ts
```diff
@@ -61,6 +61,8 @@ import {
   MessageModel,
 } from "../models/message";
 
+const ApplicationInsightsClient = ApplicationInsights.defaultClient;
```
cloudify: `defaultClient` [is null unless you call setup()](https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/applicationinsights.ts#L38) (that creates a new client)

lib/controllers/messages.ts
```diff
@@ -192,7 +194,7 @@ type IGetMessagesHandler = (
  * Returns a type safe CreateMessage handler.
  */
 export function CreateMessageHandler(
-  applicationInsightsClient: ApplicationInsights.TelemetryClient,
+  applicationInsightsClient: typeof ApplicationInsightsClient,
```
cloudify: why?

cloudify:
`EventTelemetry` [extends Telemetry](https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/Library/TelemetryTypes/Telemetry.ts) thus inheriting the custom `properties`.

cloudify:
`EventTelemetry` [extends Telemetry](https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/Library/TelemetryTypes/Telemetry.ts) thus inheriting the custom `properties`.

PR #44, opened 2017-09-15
proposal: refactor middleware function

gunzip:
this is a proposal to refactor current middleware function using async library.
it's a drop in replacement for the current implementation (all test passes).

feel free to close this and keep for historical purpose so if and when TS will implement this:
https://github.com/Microsoft/TypeScript/issues/5453 we can get rid of the overloading decls as well.


gunzip:
closing until the following gets fixed:
https://github.com/DefinitelyTyped/DefinitelyTyped/issues/16980

PR #45, opened 2017-09-18
[#151122611] Use winston for logging.

lussoluca:


cloudify:
Before using `winston` for logging, you have to initialize the custom transport that uses `context.log()`, like I did [here](https://github.com/teamdigitale/digital-citizenship-functions/blob/master/lib/public_api_v1.ts#L84)

cloudify:
Also please add a few lines of description of the logging mechanism to the `README` (imagine a contributor that wants to understand how to add logging to some new piece of code).

PR #49, opened 2017-09-18
Fixes missing check for optional attributes on generated models

cloudify:


lib/api/definitions/ProblemJson.ts
```diff
@@ -29,11 +29,11 @@ export interface ProblemJson {
 export function isProblemJson(arg: any): arg is ProblemJson {
   return (
     arg &&
-    typeof arg.type === "string" &&
-    typeof arg.title === "string" &&
-    isHttpStatusCode(arg.status) &&
-    typeof arg.detail === "string" &&
-    typeof arg.instance === "string" &&
+    (!arg.type || typeof arg.type === "string") &&
```
gunzip: !0 = true

shouldn't be arg.type === undefined ?

lib/api/definitions/PaginationResponse.ts
```diff
@@ -21,8 +21,8 @@ export interface PaginationResponse {
 export function isPaginationResponse(arg: any): arg is PaginationResponse {
   return (
     arg &&
-    typeof arg.page_size === "number" &&
-    typeof arg.next === "string" &&
+    (!arg.page_size || typeof arg.page_size === "number") &&
```
gunzip: !"" = true, see below

PR #50, opened 2017-09-19
[#151140395] Adds user ID to created messages

cloudify:


lib/utils/middlewares/azure_api_auth.ts
```diff
@@ -97,45 +102,68 @@ export function AzureApiAuthMiddleware(
...
+      if (maybeGroups.isEmpty) {
+        // the use as no valid authorization groups
+        return resolve(left(ResponseErrorForbiddenNoAuthorizationGroups));
+      }
+
+      // now we have some valid groups that the users is part of
+      const groups = maybeGroups.get;
+
+      // helper that checks whether the user is part of a specific group
+      const userHasOneGroup = (name: UserGroup) => groups.has(name);
```
gunzip: maybe it's better to put this at the module level and / or make a dedicated user module with this kind of utilities.

cloudify: not sure what you mean, how would you make this helper generic?
it's just a call the `Set#has` method on the locally scoped `groups`

PR #51, opened 2017-09-19
Generates some new models from the openapi specs

cloudify:
NotificationChannelStatus and NotificationStatus


api/public_api_v1.yaml
```diff
@@ -210,14 +209,22 @@ definitions:
         $ref: '#/definitions/TimeToLive'
       content:
         $ref: '#/definitions/MessageContent'
-      status:
-        $ref: '#/definitions/MessageStatus'
+      senderOrganizationId:
```
gunzip: why camel cased ?

api/public_api_v1.yaml
```diff
@@ -210,14 +209,22 @@ definitions:
...
       - id
       - fiscal_code
-      - time_to_live
       - content
-      - status
+      - senderOrganizationId
+  MessageStatus:
+    type: object
+    properties:
+      message:
```
gunzip: why message status contains the whole message ?

lib/controllers/messages.ts
```diff
@@ -127,6 +113,22 @@ export const MessagePayloadMiddleware: IRequestMiddleware<
...
+function retrievedMessageToPublic(
+  retrievedMessage: IRetrievedMessage
+): CreatedMessage {
+  return {
+    content: {
+      body_short: retrievedMessage.bodyShort
+    },
+    fiscal_code: retrievedMessage.fiscalCode,
+    id: retrievedMessage.id,
+    senderOrganizationId: retrievedMessage.senderOrganizationId
```
gunzip: sender_organization_id

api/public_api_v1.yaml
cloudify: ops leftover

api/public_api_v1.yaml
cloudify: better rename it, this is actually the response to `GetMessage`

PR #52, opened 2017-09-19
Generates more models from the API specs

cloudify:
CreatedMessage, CreatedMessageResponse, ExtendedProfile, LimitedProfile


api/public_api_v1.yaml
```diff
@@ -104,17 +91,18 @@ paths:
...
           schema:
             $ref: '#/definitions/ProblemJson'
-  '/preferences/{fiscal_code}':
-    x-swagger-router-controller: Preferences
+  '/profiles/{fiscal_code}':
     get:
-      operationId: getUserPreferences
+      operationId: getProfile
       summary: Get User's Preferences
       description: Returns the preferences for the user identified by the provided fiscal code.
```
gunzip: align description with the new 'profile' naming. 
moreover the description should explain when it returns Limited vs Extended profile.

lib/controllers/messages.ts
```diff
@@ -308,10 +298,15 @@ export function CreateMessageHandler(
...
 
+      const responsePayload: CreatedMessageResponse = {
+        dry_run: false
+      };
+
       // redirect the client to the message resource
       return ResponseSuccessRedirectToResource(
         retrievedMessage,
-        `/api/v1/messages/${fiscalCode}/${message.id}`
+        `/api/v1/messages/${fiscalCode}/${message.id}`,
```
gunzip: better to not hardcode API paths (expecially true for version)

lib/utils/response.ts
```diff
@@ -78,21 +78,24 @@ export function ResponseSuccessJsonIterator<T>(
...
+export function ResponseSuccessRedirectToResource<T, V>(
   resource: T,
-  url: string
-): IResponseSuccessRedirectToResource<T> {
+  url: string,
+  payload: V
+): IResponseSuccessRedirectToResource<T, V> {
   return {
-    apply: res => res.redirect(202, url),
+    apply: res => res.json(payload).redirect(202, url),
```
gunzip: why ?

api/public_api_v1.yaml
cloudify: descriptions and operations will be my next effort (contribution welcome), I'm currently focusing on generating models

lib/controllers/messages.ts
cloudify: they have to be hardcoded _somewhere_ :)

lib/utils/response.ts
cloudify: for consistency, I'd like to always return a response :)

lib/controllers/messages.ts
cloudify: also that code is not part of this PR, I'll add a reminder in the backlog

cloudify: https://www.pivotaltracker.com/story/show/151217639

PR #54, opened 2017-09-19
[#150738001] Publishes the openapi specs to /swagger.json

cloudify:


lib/controllers/openapi.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import { wrapRequestHandler } from "../utils/request_middleware";
+
+import { IResponseSuccessJson, ResponseSuccessJson } from "../utils/response";
+
+type GetOpenapi = () => Promise<IResponseSuccessJson<object>>;
+
+function getOpenapiHandler(apiSpecs: object): GetOpenapi {
+  return () => {
+    return new Promise(resolve => {
```
gunzip: maybe return Promise.resolve(..) is more concise =)

lib/controllers/openapi.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import * as express from "express";
+
+import { wrapRequestHandler } from "../utils/request_middleware";
+
+import { IResponseSuccessJson, ResponseSuccessJson } from "../utils/response";
+
+type GetOpenapi = () => Promise<IResponseSuccessJson<object>>;
+
+function getOpenapiHandler(apiSpecs: object): GetOpenapi {
```
gunzip: why not just generate the json (instead of the .ts) and serve it with express.static ?

cloudify: not super strong reasons, it comes down to let the compiler catch issues at compile time (vs runtime)

in this way we can enforce the presence (and the correctness) of the specs at compile time (vs trying to load a file at runtime that for some mistake was not included in the build) - if for some reason the specs file is not there, the build will fail

PR #56, opened 2017-09-22
[#151262556] Save message content in inbox only if enabled in profile

cloudify:
This commit also decouples the storage of message content in the message
record from the CreateMessage API handler, moving it to the async create
message handles.

### DIFF
--- a/lib/created_message_queue_handler.ts
+++ b/lib/created_message_queue_handler.ts
@@ -25,7 +25,11 @@ import {
   ICreatedMessageEvent,
   isICreatedMessageEvent
 } from "./models/created_message_event";
-import { IRetrievedMessage } from "./models/message";
+import {
+  IMessageContent,
+  IRetrievedMessageWithoutContent,
+  MessageModel
+} from "./models/message";
 import {
   INewNotification,
   INotificationChannelEmail,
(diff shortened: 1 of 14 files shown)


lib/created_message_queue_handler.ts
```diff
@@ -100,71 +110,95 @@ export async function handleMessage(
...
-        // TODO: we could check the error to see if it's actually transient
-        return left(ProcessingError.TRANSIENT);
+  // query succeeded, we may have a profile
+  const maybeProfile = errorOrMaybeProfile.right;
+
+  // whether the recipient wants us to store the message content
+  const isMessageStorageEnabled =
+    maybeProfile
+      .map(profile => profile.isStorageOfMessageContentEnabled)
+      .getOrElse(false) || false;
```
gunzip: why || false ?

lib/created_message_queue_handler.ts
```diff
@@ -100,71 +110,95 @@ export async function handleMessage(
...
+  if (maybeEmailNotification.isEmpty) {
+    // no channells to notify the user
+    return left(ProcessingError.NO_ADDRESSES);
+  }
+
+  // create a new Notification object with the configured notification channels
+  const notification: INewNotification = {
+    emailNotification: maybeEmailNotification.isDefined
+      ? maybeEmailNotification.get
+      : undefined,
```
gunzip: what it means 'undefined' here ? (maybe it's worth a comment)

lib/created_message_queue_handler.ts
cloudify: `isStorageOfMessageContentEnabled` can be `undefined`, so `getOrElse(false)` returns `false | undefined`, so: the first `false` is when there's no profile, the 2nd `false` is when there's a profile but `isStorageOfMessageContentEnabled` is undefined.

lib/created_message_queue_handler.ts
cloudify: = no email notification (there's no email address to send the notification to)

cloudify: yes it looks weird, since there's a check for `maybeEmailNotification.isEmpty` before, this is actually for later, when we have multiple channels, some of them may not be configured (thus the `undefined`)

I'd like to keep it (to avoid the risk of forgetting to add the check later)

PR #58, opened 2017-09-26
chore: [#150708172] added blob attachments

gunzip:
- replace messageModel.update() in queue handler with messageModel.attachBlob()
- added dotenv package to retrieve environment variables from local file (useful for testing)
- added utility functions to store blob(s)
- added model methods to store attachments (= associate media links to cosmosdb documents)


lib/created_message_queue_handler.ts
```diff
@@ -125,16 +125,22 @@ export async function handleMessage(
...
-    const errorOrUpdatedMessage = await messageModel.update(
+    // we add the content of the message to the blob storage for later retrieval
+    const errorOrRetrievedMessage = await messageModel.attachStoredContent(
       retrievedMessage.id,
       retrievedMessage.fiscalCode,
-      m => {
-        return { ...m, content: messageContent };
+      {
+        content: messageContent,
+        // we do not store documentdb metadata
```
cloudify: comment refers to `fiscalCode`?

lib/created_message_queue_handler.ts
```diff
@@ -125,16 +125,22 @@ export async function handleMessage(
...
     // if the recipient wants to store the messages
-    // we add the content of the message to the message record
-    const errorOrUpdatedMessage = await messageModel.update(
+    // we add the content of the message to the blob storage for later retrieval
+    const errorOrRetrievedMessage = await messageModel.attachStoredContent(
       retrievedMessage.id,
       retrievedMessage.fiscalCode,
-      m => {
-        return { ...m, content: messageContent };
+      {
```
cloudify: just storing `messageContent` should be sufficient?

i.e. a `Message` attachment ===  `MessageContent`

lib/created_message_queue_handler.ts
```diff
@@ -125,16 +125,22 @@ export async function handleMessage(
...
+        content: messageContent,
+        // we do not store documentdb metadata
+        fiscalCode: retrievedMessage.fiscalCode,
+        senderOrganizationId: retrievedMessage.senderOrganizationId,
+        senderUserId: retrievedMessage.senderUserId
       }
     );
 
-    if (errorOrUpdatedMessage.isLeft) {
+    winston.debug(`storingMessageContent|${JSON.stringify(retrievedMessage)}`);
```
cloudify: if you want to follow my convention, first identifier is the method name, i.e. `handleMessage`

lib/models/message.ts
```diff
@@ -281,4 +289,45 @@ export class MessageModel extends DocumentDbModel<
...
+
+  public async attachStoredContent(
+    messageId: string,
+    partitionKey: string,
+    message: IMessageWithContent
+  ): Promise<Either<Error, Option<IRetrievedMessage>>> {
+    const blobId = this.getMessageAttachmentName(messageId);
+
+    // store media (attachment) with message content in blob storage
+    const errorOrStoredMessage = await upsertBlobFromText(
```
cloudify: Using just `Message` may be confusing (in other parts of the code `Message` is just the metadata), better refer to `MessageContent` in this case: `errorOrMessageContent`

lib/utils/documentdb_model.ts
```diff
@@ -153,4 +153,46 @@ export abstract class DocumentDbModel<
...
+    }
+
+    // return the document with attachment media information
+    return this.find(documentId, partitionKey);
+  }
+
+  public async getAttachments(
+    documentId: string,
+    options: DocumentDb.FeedOptions = {}
+  ): Promise<ReadonlyArray<DocumentDb.AttachmentMeta>> {
```
cloudify: I would return an Iterator here and let the client decide whether to fetch everything with `iteratorToArray` or lazily process the results.

lib/utils/storage.ts
```diff
@@ -0,0 +1,61 @@
+import * as azure from "azure-storage";
```
cloudify: since this is specifically for Azure Storage, better call the file `azure_storage.ts`

lib/utils/storage.ts
```diff
@@ -0,0 +1,61 @@
...
+import { Option, option } from "ts-option";
+import { Either, left, right } from "./either";
+
+const STORAGE_CONNECTION_STRING = process.env.AzureWebJobsStorage;
+
+const blobService = azure.createBlobService(STORAGE_CONNECTION_STRING);
```
cloudify: you should never instantiate types (esp. services) in utility modules, let the client do that and pass the instance to the utility

lib/utils/storage.ts
```diff
@@ -0,0 +1,61 @@
...
+): string {
+  return blobService.getUrl(containerName, attachmentName);
+}
+
+/**
+ * Create a new storage container if not exists.
+ * 
+ * @param containerName 
+ */
+export function createContainerIfNotExists(
```
cloudify: this method is never used (and should never be used), pls remove it

lib/utils/storage.ts
```diff
@@ -0,0 +1,61 @@
...
+import { Either, left, right } from "./either";
+
+const STORAGE_CONNECTION_STRING = process.env.AzureWebJobsStorage;
+
+const blobService = azure.createBlobService(STORAGE_CONNECTION_STRING);
+
+export function getBlobUrl(
+  containerName: string,
+  attachmentName: string
+): string {
```
cloudify: this is a good opportunity to use tagged types instead of plain strings (e.g. `DocumentDbAttachmentUrl`)

lib/utils/storage.ts
```diff
@@ -0,0 +1,61 @@
...
+        resolve(right(option(result)));
+      }
+    });
+  });
+}
+
+/**
+ * Create a new blob (media). 
+ * Assumes that the container <containerName> already exists.
+ *
```
cloudify: remove extra empty line

lib/utils/storage.ts
```diff
@@ -0,0 +1,61 @@
+import * as azure from "azure-storage";
+import { Option, option } from "ts-option";
+import { Either, left, right } from "./either";
+
+const STORAGE_CONNECTION_STRING = process.env.AzureWebJobsStorage;
```
cloudify: only the main function should have access to the environment, utility functions should be "pure"

lib/created_message_queue_handler.ts
gunzip: it may be sufficient but there's a case for storing the whole message (fiscalCode etc.): if we will implement an azure elasticsearch task running on the whole archive, it would be better to have all the information in the same blob. 

otherwise, we can just store the message content.

lib/created_message_queue_handler.ts
gunzip: no, to documentdb fields (_self, etc.)

lib/created_message_queue_handler.ts
gunzip: ok

lib/models/message.ts
gunzip: ok

lib/utils/documentdb_model.ts
gunzip: ok

lib/utils/storage.ts
gunzip: ok

lib/utils/storage.ts
gunzip: ok

lib/utils/documentdb_model.ts
gunzip: ok

gunzip:
rebased into another branch

PR #59, opened 2017-09-26
feat(messaging): [#150708172] store message content in blob storage

gunzip:


README.md
```diff
@@ -118,7 +118,7 @@ The high level flow of the data is the following.
       preferences for the `Profile` associated to the recipient of the `Message`
         and create a pending `Notification`. If the user enabled the
         _message inbox_, the content of the `Message` will also be persisted
-        and associated to the `Message` record.
+        and associated to the `Message` record in a blob container named 'message-content'.
```
cloudify: use backticks for identifiers -> `message-content`

lib/created_message_queue_handler.ts
```diff
@@ -102,7 +106,8 @@ export async function handleMessage(
   notificationModel: NotificationModel,
   retrievedMessage: IRetrievedMessageWithoutContent,
   messageContent: IMessageContent,
-  defaultAddresses: Option<NewMessageDefaultAddresses>
+  defaultAddresses: Option<NewMessageDefaultAddresses>,
+  blobService: BlobService
```
cloudify: it's good practice to keep optional parameters at the end of param list, also as a convention I usually first keep the "services" first, so I would place `blobService` right after or before `notificationModel`.

lib/created_message_queue_handler.ts
```diff
@@ -125,16 +130,23 @@ export async function handleMessage(
...
-    // we add the content of the message to the message record
-    const errorOrUpdatedMessage = await messageModel.update(
+    // we add the content of the message to the blob storage for later retrieval
+    const errorOrAttachment = await messageModel.attachStoredContent(
+      blobService,
       retrievedMessage.id,
       retrievedMessage.fiscalCode,
-      m => {
-        return { ...m, content: messageContent };
+      {
```
cloudify: please just store the `messageContent`, the metadata is already stored in the message (we can look it up by ID) - besides, I would not consider yet `senderOrganizationId` and `senderUserId` as "consolidated" metadata, we may need/want to change their meaning before GA.

```javascript
attachStoredContent(
  blobService,
  retrievedMessage.id,
  retrievedMessage.fiscalCode,
  messageContent
);
```

lib/models/__tests__/message.test.ts
```diff
@@ -286,3 +292,51 @@ describe("findMessageForRecipient", () => {
     }
   });
 });
+
+describe("attachStoredContent", () => {
+  const aMessageId = "MESSAGE_ID";
+  const aPartitionKey = "PARTITION_KEY";
+  const anAttachmentMeta = { name: "attachmentMeta" };
+  const clientMock = {
```
cloudify: You should instantiate this inside the test as the `upsertAttachment` mock will be not be reset between tests.

As an alternative you can call [resetMock](https://facebook.github.io/jest/docs/en/mock-function-api.html#mockfnmockreset) on `upsertAttachment` on each test.

lib/models/message.ts
```diff
@@ -5,12 +5,21 @@ import * as DocumentDbUtils from "../utils/documentdb";
...
 import { BodyShort, isBodyShort } from "../api/definitions/BodyShort";
 import { FiscalCode, isFiscalCode } from "../api/definitions/FiscalCode";
 
+import {
+  BlobService,
+  getBlobUrl,
+  upsertBlobFromText
+} from "../utils/azure_storage";
+
+const MESSAGE_BLOB_STORAGE_CONTAINER_NAME = "message-content";
```
cloudify: I would pass the name of the container to the `Message` model constructor as a parameter and in turn read the value from the environment from the code that instantiates the `Message` model.

lib/models/message.ts
```diff
@@ -281,4 +293,52 @@ export class MessageModel extends DocumentDbModel<
       query: "SELECT * FROM messages m WHERE (m.fiscalCode = @fiscalCode)"
     });
   }
+
+  public async attachStoredContent(
```
cloudify: add a short comment on the purpose of the function

lib/models/message.ts
```diff
@@ -281,4 +293,52 @@ export class MessageModel extends DocumentDbModel<
...
+          `Error while attaching stored message: ${errorOrAttachmentMeta.left
+            .code} - ${errorOrAttachmentMeta.left.body}`
+        )
+      );
+    }
+
+    return right(errorOrAttachmentMeta.right);
+  }
+
+  private getMessageAttachmentName(id: string): string {
```
cloudify: this can be inlined since it's called only once, also it will make it more clear what is going on

lib/models/message.ts
```diff
@@ -281,4 +293,52 @@ export class MessageModel extends DocumentDbModel<
...
     });
   }
+
+  public async attachStoredContent(
+    blobService: BlobService,
+    messageId: string,
+    partitionKey: string,
+    message: IMessageWithContent
+  ): Promise<Either<Error, Option<DocumentDb.AttachmentMeta>>> {
+    // this is the attachment id __and__ the filename
```
cloudify: comment is not very clear, I believe is safe to inline `getMessageAttachmentName` here since it's called only from here

lib/models/message.ts
```diff
@@ -281,4 +293,52 @@ export class MessageModel extends DocumentDbModel<
...
+    const blobId = this.getMessageAttachmentName(messageId);
+
+    // store media (attachment) with message content in blob storage
+    const errorOrMessageContent = await upsertBlobFromText(
+      blobService,
+      MESSAGE_BLOB_STORAGE_CONTAINER_NAME,
+      blobId,
+      JSON.stringify({ id: messageId, ...message })
```
cloudify: is the following getting stored in the blob? (full message + its ID) ? we should just store the `content`, that's it. 

```json
{
  "id": "...",
  "message": {
    ...,
    "content": {
      ...
    }
  }
}
```

lib/created_message_queue_handler.ts
cloudify: I see now (after looking at the implementation of `attachStoredContent`), you're passing an `IMessageWithContent` - we just need to store the content, not the whole message!

lib/models/message.ts
```diff
@@ -281,4 +293,52 @@ export class MessageModel extends DocumentDbModel<
...
+    );
+
+    if (errorOrMessageContent.isLeft) {
+      return left(errorOrMessageContent.left);
+    }
+
+    // attach the created media to the message identified by messageId and partitionKey
```
cloudify: `media` -> `attachment`

lib/models/message.ts
```diff
@@ -281,4 +293,52 @@ export class MessageModel extends DocumentDbModel<
...
+      media: getBlobUrl(
+        blobService,
+        MESSAGE_BLOB_STORAGE_CONTAINER_NAME,
+        blobId
+      )
+    });
+
+    if (errorOrAttachmentMeta.isLeft) {
+      return left(
+        new Error(
```
cloudify: If you don't want to lose the `QueryError` (by converting it to an `Error`), you could change the return type of this function to:

`Promise<Either<Error | QueryError, Option<DocumentDb.AttachmentMeta>>>`

or

`Promise<Either<Either<Error, QueryError>, Option<DocumentDb.AttachmentMeta>>>`

lib/utils/__tests__/azure_storage.test.ts
```diff
@@ -0,0 +1,64 @@
...
+  requestId: ""
+};
+const aConnectionString = "credentials=credentials";
+const aContainerName = "container-name";
+const anAttachmentName = "attachment-name";
+const aRandomText = "random-text";
+
+const aBlobService = new azureStorage.BlobService(aConnectionString);
+
+// function flushPromises<T>(): Promise<T> {
```
cloudify: the commented out function can be removed

lib/utils/azure_storage.ts
```diff
@@ -0,0 +1,48 @@
+import * as azureStorage from "azure-storage";
```
cloudify: would be nice a block comment explaining what this file provides (e.g. helpers for the Azure Storage service)

lib/utils/azure_storage.ts
```diff
@@ -0,0 +1,48 @@
...
+import { Option, option } from "ts-option";
+import { Either, left, right } from "./either";
+
+// Storage connection string may be process.env.AzureWebJobsStorage;
```
cloudify: ?

lib/utils/azure_storage.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export type BlobService = azureStorage.BlobService;
```
cloudify: I'm not very fond of aliasing and re-exporting a type from another library, is this strictly required or it is just to save an `import`?

lib/utils/azure_storage.ts
```diff
@@ -0,0 +1,48 @@
...
+export type BlobResult = azureStorage.BlobService.BlobResult;
+
+export function getBlobService(connectionString: string): BlobService {
```
cloudify: what value this function adds? it just calls `azureStorage.createBlobService` with the same params, this can be written as:

`export const getBlobService  = azureStorage.createBlobService;`

(same comment as above, why is it needed?)

lib/utils/azure_storage.ts
```diff
@@ -0,0 +1,48 @@
...
+  return azureStorage.createBlobService(connectionString);
+}
+
+export function getBlobUrl(
+  blobService: BlobService,
+  containerName: string,
+  attachmentName: string
+): string {
+  return blobService.getUrl(containerName, attachmentName);
```
cloudify: ditto!

helper functions should add some value, be convenient in some way!

lib/utils/azure_storage.ts
```diff
@@ -0,0 +1,48 @@
...
+): Promise<Either<Error, Option<BlobResult>>> {
+  return new Promise((resolve, _) =>
+    blobService.createBlockBlobFromText(
+      containerName,
+      blobName,
+      text,
+      (err, result, __) => {
+        if (err) {
+          resolve(left(err));
+        }
```
cloudify: need an `else` here or a `return resolve(...)` above

or else in case of error, when the first `resolve` returns the 2nd resolve will be called

gunzip: Good catch

lib/created_message_queue_handler.ts
gunzip: Ok

lib/created_message_queue_handler.ts
gunzip: Ok

lib/models/message.ts
gunzip: actually media and attachment are two different thigs (expecially inside azure documentation).

a media is a blob stored inside azure blob storage. an attachment is some metadata associated to the cosmosdb document that link the media to the document.

so an attachment cannot be stored inside blob storage, as it live inside cosmosdb as metadata.

lib/models/message.ts
gunzip: ok

lib/models/message.ts
gunzip: ok

lib/models/message.ts
gunzip: ok

lib/models/__tests__/message.test.ts
gunzip: ok

lib/models/message.ts
gunzip: ok

lib/utils/__tests__/azure_storage.test.ts
gunzip: ok

lib/utils/azure_storage.ts
gunzip: ok

lib/utils/azure_storage.ts
gunzip: short answer: yes, that's exactly to save an import.

long answer: the original intent of this file was to abstract away some ops on storage (that's why was called just 'storage' and not 'azure_storage'). so these type aliases and the helper methods were needed to encapsulate "swappable" types and behaviors.

not that the semantic has changed and it's bound to azure, i agree that we can get rid of type aliases and helper methods.

lib/utils/azure_storage.ts
gunzip: see https://github.com/teamdigitale/digital-citizenship-functions/pull/59#discussion_r141296878

lib/utils/azure_storage.ts
gunzip: see https://github.com/teamdigitale/digital-citizenship-functions/pull/59#discussion_r141296878

lib/models/message.ts
gunzip: see https://github.com/teamdigitale/digital-citizenship-functions/pull/59#discussion_r141296878

lib/created_message_queue_handler.ts
```diff
@@ -40,14 +42,17 @@ import {
...
-import { toNonEmptyString } from "./utils/strings";
+import { NonEmptyString, toNonEmptyString } from "./utils/strings";
 import { Tuple2 } from "./utils/tuples";
 
 // Setup DocumentDB
 
 const COSMOSDB_URI: string = process.env.CUSTOMCONNSTR_COSMOSDB_URI;
 const COSMOSDB_KEY: string = process.env.CUSTOMCONNSTR_COSMOSDB_KEY;
 
+const MESSAGE_CONTAINER_NAME: NonEmptyString =
```
cloudify: what you get from the `env` is a `string`, you must properly convert it to `NonEmptyString` if you want that (either with `isNonEmptyString` or with `toNonEmptyString`)

lib/public_api_v1.ts
```diff
@@ -38,6 +39,9 @@ const app = express();
 const COSMOSDB_URI: string = process.env.CUSTOMCONNSTR_COSMOSDB_URI;
 const COSMOSDB_KEY: string = process.env.CUSTOMCONNSTR_COSMOSDB_KEY;
 
+const MESSAGE_CONTAINER_NAME: NonEmptyString =
```
cloudify: same here you don't know whether `process.env.MESSAGE_CONTAINER_NAME` is empty or not, you should go through `isNonEmptyString` or `toNonEmptyString`

lib/created_message_queue_handler.ts
gunzip: ok that's a more general issue. we don't know if process.env variables are populated till they're used. 

if you use toNonEmptyString() here, when you'll import the module in tests, you'll MUST set the environment variable (even if you use mocks) before importing or the test will throw an exeception calling get().

said that, probably is better here to just use a string instead of NonEmptyString for MESSAGE_CONTAINER_NAME. what about it ?

PR #60, opened 2017-09-27
[#150738459] Adds functionality for formatting message content

cloudify:
This is the first of a multi-PR implementation.

This first PR changes `MessageContent` to accept a subject and a content in Markdown format (no validation happens yet). The Markdown gets then converted to HTML and used as the body of the email notification.


api/public_api_v1.yaml
```diff
@@ -223,18 +222,22 @@ definitions:
...
+  MessageBodyMarkdown:
     type: string
-    description: 'A long version of the message, in plain text. The content of this field will be delivered to channels that don''t have any limit in terms of content size (e.g. email, etc...). The long version of the message is optional and should be provided only in cases when a longer message (compared to the short version provided in body_short) gives more information to the user. This field should not be a duplicate of body_short.'
+    description: |-
+      The full version of the message, in plain text or Markdown format. The
+      content of this field will be delivered to channels that don't have any
+      limit in terms of content size (e.g. email, etc...).
     minLength: 80
-    maxLength: 8000
+    maxLength: 10000
```
gunzip: can we predict the message size after markdown parsing ? that's because actually we are sending the whole message to queue channels (max 64kb).  is it possible that the generated html has a bigger size so that the payload won't fit in the queue message ?

lib/emailnotifications_queue_handler.ts
```diff
@@ -173,6 +169,12 @@ async function handleNotification(
     return left(ProcessingError.PERMANENT);
   }
 
+  // converts the markdown body to HTML
+  // TODO: handle errors / validate the markdown
+  const bodyHtml = (await markdownToHtml.process(
```
gunzip: I guess that if the message is plain/text we should skip markdown parsing (or the intended result could be different from the sender intention).

lib/emailnotifications_queue_handler.ts
```diff
@@ -182,12 +184,12 @@ async function handleNotification(
...
       "X-Italia-Messages-MessageId": messageId,
-      "X-Italia-Messages-NotificationId": notification.id
+      "X-Italia-Messages-NotificationId": notificationId
     },
-    html: messageContent.bodyShort,
+    html: bodyHtml,
     messageId,
     subject: "Un nuovo avviso per te.",
-    text: messageContent.bodyShort,
+    // text: messageContent.bodyMarkdown,
```
gunzip: how to handle this one ?

lib/emailnotifications_queue_handler.ts
```diff
@@ -198,12 +200,12 @@ async function handleNotification(
   const eventContent = {
     addressSource: emailNotification.addressSource,
     messageId,
-    mta: "sendgrid",
-    notificationId
+    notificationId,
+    transport: "sendgrid"
```
gunzip: can we made the transport configurable ?

package.json
```diff
@@ -38,11 +39,16 @@
     "js-yaml": "^3.10.0",
     "nodemailer": "^4.0.1",
     "nodemailer-sendgrid-transport": "^0.2.0",
+    "rehype-stringify": "^3.0.0",
```
gunzip: that's a lot of stuff to just convert markdown to html (there are simpler packages like marked). 
that would be fine anyway if we'll need this tools for more complex processing in the near future.

cloudify: yes I selected rehype because it will make it easier to pre-process and post-process the content (needed for https://www.pivotaltracker.com/story/show/151434506 and https://www.pivotaltracker.com/story/show/151284725)

lib/emailnotifications_queue_handler.ts
cloudify: anyway, I'm thinking about providing an optional `text` attribute in `MessageContent` but until now all the examples we got from the SAS have all rich content, so not sure how much value could we have in that use case - we will see, it's not going to be too much effort to add a plain text option

lib/emailnotifications_queue_handler.ts
cloudify: I was considering two options:

* processing the generated HTML with [this](https://www.npmjs.com/package/html-to-text)
* strip the Markdown tokens with something like [this](https://www.npmjs.com/package/remove-markdown)

I'm currently inclined to implement the first option as it will do nice things like re-flowing the text to 80 columns and formatting tables.

lib/emailnotifications_queue_handler.ts
cloudify: you mean the actual transport used by nodemailer or the value of the attribute in the event?

PR #65, opened 2017-09-28
[#151512955] Wraps email content into an HTML template

cloudify:
Sono partito da [questo template](https://mjml.io/try-it-live/templates/card).

Poi sarà da sistemare un po' e/o farne altri.


Makefile
```diff
@@ -0,0 +1,30 @@
+#
```
gunzip: for easier building steps (read: don't require make) and cross-platform compatibility (ie. to run this locally), can we convert this Makefile into a  shelljs (node) script ?

lib/__tests__/emailnotifications_queue_handler.test.ts
```diff
@@ -167,7 +167,9 @@ describe("handleNotification", () => {
     expect(sentMail.data.headers["X-Italia-Messages-NotificationId"]).toBe(
       "A_NOTIFICATION_ID"
     );
-    expect(sentMail.data.html).toBe("<h1>Hello world!</h1>");
+    expect(
+      String(sentMail.data.html).indexOf("<h1>Hello world!</h1>")
+    ).toBeGreaterThan(0);
```
gunzip: maybe it's greater or equal ?

lib/emailnotifications_queue_handler.ts
```diff
@@ -186,14 +215,22 @@ export async function handleNotification(
...
-  // converts the markdown body to HTML
-  // TODO: handle errors / validate the markdown
-  const bodyHtml = (await markdownToHtml.process(
+  const documentHtmlOrError = await generateDocumentHtml(
+    subject,
     messageContent.bodyMarkdown
-  )).toString();
+  );
+
+  if (documentHtmlOrError.isLeft) {
```
gunzip: as a broader issue, we should remember to consider that with all this message processing there are a lot of possible failure reasons (a lot more than when sending plain text). 

the parser(s) must be resilient to errors or we risk to miss a lot of targets.

lib/templates/html/default.ts
```diff
@@ -0,0 +1,102 @@
+// DO NOT EDIT THIS FILE
```
gunzip: maybe better to not commit this file a make a npm task at build time to generate it ?

Makefile
cloudify: What build system doesn't have `make` installed? besides, `make` is a build system, not a shell, we may replace it with something like `gulp` instead (much more code to write though). Feel free to add a story for making the build process work on Windows :D

lib/__tests__/emailnotifications_queue_handler.test.ts
cloudify: Greater because there's supposed to be a `document` section before, the HTML shouldn't start with `<h1>`

lib/emailnotifications_queue_handler.ts
cloudify: indeed, that's why we log errors and we will monitor the system closely

cloudify: indeed, that's why we log errors and we will monitor the system closely

cloudify: indeed, that's why we log errors and we will monitor the system closely

lib/templates/html/default.ts
cloudify: what's the benefit in not committing it?

Makefile
gunzip: actually it works on windows (that's my developement environment =)

lib/templates/html/default.ts
gunzip: - does not appear in diffs
- it's not immediately clear if it's aligned to the lastest template

PR #68, opened 2017-09-29
[#151365021] API Definition Templates tests

gmereu:


lib/api/definitions/__tests__/MessageBodyMarkdown.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  test("isMessageBodyMarkdown on string shorter then 80 characters", () => {
+    // String shorter than 80 characters.
+    const text: string = "Lorem ipsum dolor sit amet";
+    expect(isMessageBodyMarkdown(text)).toBe(false);
+  });
+
+  test("isMessageBodyMarkdown on string longer then 10000 characters", () => {
+    // String long 90 characters.
+    const text: string =
```
gunzip: maybe we should consider adding a libray like faker to generate fixtures
https://github.com/marak/Faker.js/

lib/api/definitions/__tests__/CreatedMessage.test.ts
```diff
@@ -0,0 +1,261 @@
...
+
+import { MessageContent } from "../MessageContent";
+
+import { toMessageSubject } from "../MessageSubject";
+
+import { toFiscalCode } from "../FiscalCode";
+
+import { toTimeToLive } from "../TimeToLive";
+
+describe("Check CreatedMessage methods", () => {
```
cloudify: you should have a `describe` block for each method (you're testing each method)

then you should have a `test()` or `it()` block for each test you want to execute

for instance

```
describe("CreatedMessage#toCreatedMessage", () => {
  it("should return a defined option for valid payloads", () => {
  });

  it("should return an empty option for invalid payloads", () => {
  });
});
```

cloudify:
@gmereu come procede? :)

gmereu:
@cloudify ho revisionato la metà dei test. tra oggi e domani mattina dovrei avere terminato.

gmereu:
@cloudify ho terminato la correzione dei test secondo le tue indicazioni.

lib/api/definitions/__tests__/CreatedMessage.test.ts
```diff
@@ -0,0 +1,349 @@
...
+import { MessageContent } from "../MessageContent";
+
+import { toMessageSubject } from "../MessageSubject";
+
+import { toFiscalCode } from "../FiscalCode";
+
+import { toTimeToLive } from "../TimeToLive";
+
+describe("CreatedMessage#toCreatedMessage", () => {
+  test("should returns a defined option for valid payloads", () => {
```
cloudify: you should use `it` instead of `test` like the rest of the tests

cloudify: `returns` -> `return`

lib/api/definitions/__tests__/CreatedMessage.test.ts
```diff
@@ -0,0 +1,349 @@
...
+    const messageOne = {
+      fiscal_code: toFiscalCode("AAABBB01C01A000A").get,
+      id: "12345",
+      sender_organization_id: "Sender Organization",
+      time_to_live: toTimeToLive(3600).get
+    };
+    expect(isCreatedMessage(messageOne)).toBe(true);
+  });
+  test("should returns true if CreatedMessage object does have content property set to null", () => {
+    /* tslint:disable */
```
cloudify: do not disable tslint altogether, instead just disable the specific rules you need to disable

lib/api/definitions/__tests__/CreatedMessage.test.ts
```diff
@@ -0,0 +1,349 @@
...
+
+    const message: CreatedMessage = {
+      content: messageContent,
+      fiscal_code: toFiscalCode("AAABBB01C01A000A").get,
+      id: "12345",
+      sender_organization_id: "Sender Organization",
+      time_to_live: toTimeToLive(3600).get
+    };
+
+    expect(toCreatedMessage(message).get).toEqual(message);
```
cloudify: what's the point here? `message` the compiler already checks that `message` is a `CreatedMessage` - I think you should change your testing strategy here, what you want to do is to generate a number of fixtures for valid and invalid objects (either manually like [we do here](https://github.com/teamdigitale/digital-citizenship-functions/blob/master/lib/models/__tests__/notification_event.test.ts#L43) or automatically as per @gunzip suggestion)

gmereu:
@cloudify ho corretto i test seguendo il modello che i hai linkato.

cloudify:
I'm closing this PR since the work is stalled

PR #69, opened 2017-09-29
chore: convert makefile to gulp

gunzip:


gulpfile.js
```diff
@@ -0,0 +1,43 @@
...
+const TYPESCRIPT_TEMPLATES_DIR = "lib/templates/html";
+const MJML_TEMPLATES_DIR = "templates/mjml";
+const GIT_COMMIT_HASH = shell.exec("git rev-parse --short HEAD");
+
+const toMjml = (content, options) => {
+  const name = path.basename(options.sourcePath);
+  return [
+    `// DO NOT EDIT THIS FILE`,
+    `// this file was auto generated from '${name}'`,
+    `// commit ${GIT_COMMIT_HASH}`,
```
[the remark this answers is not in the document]
gunzip: removed. with this PR the template is generate at the build time.

gulpfile.js
```diff
@@ -0,0 +1,41 @@
+const gulp = require("gulp");
```
cloudify: a little comment explaining what this gulp file purpose is? assume that somebody opens this file and ask you: "What is this file for?"

gulpfile.js
```diff
@@ -0,0 +1,41 @@
...
+    `  footerHtml: string`,
+    `): string {`,
+    `  return \``,
+    `${mjml.mjml2html(content, { minify: true }).html}\`;`,
+    `}`,
+    ""
+  ].join("\n");
+};
+
+gulp.task("mjml", () => {
```
cloudify: I would call the task with a name that is independent from the underlying templating system used (we could change in the future how we define/generate templates but the task name should not change). Something like `gulp templates` would do :)

PR #73, opened 2017-10-04
fix deploy script to include dev deps

gunzip:
with the current configuration deploy fails because npm installs only production deps (but we need dev deps as well).


cloudify:
We should actually not run the build on the production machines, see:

* https://www.pivotaltracker.com/story/show/151370067
* https://www.pivotaltracker.com/story/show/150764052

cloudify:
I think the issue is that `tsconfig` builds also the tests (that's why it fails since on `production` mode `jest` doesn't get installed).

Unfortunately by excluding the tests from `tsconfig`, `tsc` will never try to compile the tests, making it very difficult to catch errors in tests during development (I haven't found a way to override the `exclude` entry in `tsconfig` from the command line).

I suspect that the optimal solution would be to have a `build-production` and a `build-development` tasks that run `tsc` with two different `tsconfig` (not too bad since you can extend one config from the other, overriding specific attributes, see the [extend](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html) keyword)

cloudify:
Merging this while we work on [the new deployment process](https://www.pivotaltracker.com/story/show/151370067)

PR #75, opened 2017-10-08
 chore: #151157822 API must expose secure HTTP headers

gunzip:
added helmetjs middleware


lib/public_api_v1.ts
```diff
@@ -30,10 +30,42 @@ import { CreateMessage, GetMessage, GetMessages } from "./controllers/messages";
...
+  csp({
+    directives: {
+      defaultSrc: ["'none'"],
+      upgradeInsecureRequests: true
+    }
+  })
+);
+
+// Set up the following HTTP headers:
+//    strict-transport-security: max-age=15552000; includeSubDomains
```
cloudify: since the actual headers may change with a new version, it makes sense to link to the relevant documentation

PR #78, opened 2017-10-16
 chore: [#151621342] align swagger specs to API code

gunzip:
- removed API prefix in routes (/api/v1). manage it through API manager.
- aligned endpoints to routes


api/public_api_v1.yaml
```diff
@@ -4,40 +4,48 @@ info:
...
-      - name: fiscal_code
-        in: path
-        type: string
-        required: true
+    - "$ref": "#/parameters/PaginationRequest"
+    - name: fiscal_code
+      in: path
+      type: string
+      required: true
         description: 'The fiscal code of the user, all upper case.'
```
cloudify: there's an extra indentation here, `description`, `maxLength`, `minLength` and `pattern` must be aligned with `required`

api/public_api_v1.yaml
```diff
@@ -83,36 +91,37 @@ paths:
...
-        required: true
-        pattern: '[A-Z]{6}[0-9LMNPQRSTUV]{2}[ABCDEHLMPRST][0-9LMNPQRSTUV]{2}[A-Z][0-9LMNPQRSTUV]{3}[A-Z]'
-        maxLength: 16
-        minLength: 16
-        description: 'User''s fiscal code, all upper case.'
+    - name: fiscal_code
+      in: path
+      type: string
+      required: true
+      pattern: "[A-Z]{6}[0-9LMNPQRSTUV]{2}[ABCDEHLMPRST][0-9LMNPQRSTUV]{2}[A-Z][0-9LMNPQRSTUV]{3}[A-Z]"
```
cloudify: is it possible to reference `#/definitions/FiscalCode` here?

api/public_api_v1.yaml
cloudify: besides, is it possible to reference `#/definitions/FiscalCode` here?

api/public_api_v1.yaml
```diff
@@ -4,40 +4,48 @@ info:
...
+  "/messages/{fiscal_code}/{id}":
     parameters:
-      - name: id
-        in: path
-        type: string
-        required: true
-        description: The ID of the message.
+    - name: fiscal_code
+      in: path
+      type: string
```
cloudify: is it possible to reference `#/definitions/FiscalCode` here?

api/public_api_v1.yaml
```diff
@@ -237,33 +246,45 @@ definitions:
         description: Number of items returned for each page.
       next:
         type: string
-        description: 'Contains an URL to GET the next #<page_size> results in the retrieved collection of items.'
+        description: 'Contains an URL to GET the next #<page_size> results in the
```
cloudify: is the quote required? I would remove it for consistency with the rest

api/public_api_v1.yaml
```diff
@@ -237,33 +246,45 @@ definitions:
...
     title: A citizen's profile
-    description: Describes the citizen's profile, mostly interesting for preferences attributes.
+    description: Describes the citizen's profile, mostly interesting for preferences
+      attributes.
     type: object
     properties:
       preferred_languages:
-        description: 'Indicates the User''s preferred written or spoken language. Generally used for selecting a localized User interface. Valid values are concatenation of the ISO 639-1 two letter language code, an underscore, and the ISO 3166-1 2 letter country code; e.g., ''en_US'' specifies the language English and country US.'
-        $ref: "#/definitions/PreferredLanguages"
+        description: Indicates the User's preferred written or spoken language. Generally
```
cloudify: can the description be refactored into `#/definitions/PreferredLanguages` to avoid duplication?

cloudify: Also since this is an array, the description should read `Indicates the User's preferred written or spoken languages in order of preference.`

lib/public_api_v1.ts
```diff
@@ -118,23 +118,23 @@ const appInsightsClient = new ApplicationInsights.TelemetryClient();
 // Setup handlers
 
 const debugHandler = GetDebug(organizationModel);
-app.get("/api/v1/debug", debugHandler);
-app.post("/api/v1/debug", debugHandler);
+app.get("/debug", debugHandler);
```
cloudify: you should also update [function.json](https://github.com/teamdigitale/digital-citizenship-functions/blob/master/PublicApiV1/function.json)

cloudify: anyway, I'm not much in agreement about this, since it prevents having another HTTP Function handling the v2 API within this application (e.g. a `public_api_v2.ts`) - I see more downsides than upsides (what is the upside anyway?)

gunzip: upside is you can plug multiple versions of the API in the developer portal to test it inside the UI (but I'm not particularly biased). if we want to maintain the prefix maybe we should do this for the swagger.json endpoint as well ?

    app.get("/swagger.json", GetOpenapi(publicApiV1Specs));

api/public_api_v1.yaml
gunzip: According to [OpenAPI 2.0 schema](https://github.com/OAI/OpenAPI-Specification/blob/master/schemas/v2.0/schema.json#L810) you cannot use "$ref" pointers to #/definitions/* with parameters but you can create a $ref pointer to #/parameters/*. Moreover these parameters defintions cannot contain $ref pointers to #/definitions so you end up to have two separate (but identical) definitions for FiscalCode in the best case. That's what I'm going to plug into the specs anyway.

api/public_api_v1.yaml
gunzip: it's required as the character "#" starts a yaml comment

PR #85, opened 2017-10-24
Feat: Organization to service

gunzip:
avoid using x-user-note and provide an API for services (metadata) tied to users' subscriptions


AdminApi/function.json
```diff
@@ -0,0 +1,18 @@
+{
+  "scriptFile": "index.js",
```
cloudify: this needs to reference `../lib/index.js`

AdminApi/function.json
```diff
@@ -0,0 +1,18 @@
...
+  "disabled": false,
+  "bindings": [
+    {
+      "authLevel" : "function",
+      "type"      : "httpTrigger",
+      "direction" : "in",
+      "name"      : "req",
+      "methods"   : [ "get", "post", "put" ],
+      "route"     : "adm/v1/{*segments}"
```
cloudify: since this is an API for internal use, versioning is not strictly necessary, I would just publish it as `/admin/...`

AdminApi/index.ts
```diff
@@ -0,0 +1,4 @@
+import { index } from "../lib/admin_api";
```
cloudify: this file has been replaced by `/lib/index.ts`

lib/utils/express.ts
```diff
@@ -0,0 +1,38 @@
+import * as express from "express";
+import * as helmet from "helmet";
+import * as csp from "helmet-csp";
+import * as referrerPolicy from "referrer-policy";
+
+export function getSecureExpressApp(): express.Express {
+  const app = express();
```
cloudify: I would keep the responsibility of creating an express app outside of this function and instead have this function just configure the app with the security plugins

lib/admin_api.ts
```diff
@@ -0,0 +1,66 @@
...
+import { ServiceModel } from "./models/service";
+
+import { CreateService, UpdateService } from "./controllers/admin_services";
+
+import { GetDebug } from "./controllers/debug";
+
+import { getSecureExpressApp } from "./utils/express";
+
+// Setup Express
+const app = getSecureExpressApp();
```
cloudify: you should create an `express` instance in here and pass it to this method to configure it with the security middlewares

lib/controllers/admin_services.ts
```diff
@@ -0,0 +1,217 @@
+/*
```
cloudify: In RESTful APIS, a good convention is to have the controller named after the resource it is responsible for, in case of scoped resources (i.e. `admin/*`) a good convention is to keep the controllers under the scope folder (i.e. `lib/controllers/admin/*`)

lib/controllers/admin_services.ts
```diff
@@ -0,0 +1,217 @@
...
+  | IResponseSuccessJson<IRetrievedService>
+  | IResponseErrorValidation
+  | IResponseErrorQuery
+  | IResponseErrorInternal
+>;
+
+type IUpdateServiceHandler = (
+  auth: IAzureApiAuthorization,
+  serviceId: NonEmptyString,
+  service: IRetrievedService
```
cloudify: this should probably be a `INewService` ?

lib/controllers/admin_services.ts
```diff
@@ -0,0 +1,217 @@
...
+/**
+ * A middleware that extracts a Service payload from a request.
+ *
+ * TODO: validate the payload against a schema.
+ */
+const ServicePayloadMiddleware: IRequestMiddleware<
+  IResponseErrorValidation,
+  IService
+> = request => {
+  const servicePayload: IService = {
```
cloudify: `const body = request.body;` will save some keystrokes :)

lib/controllers/admin_services.ts
```diff
@@ -0,0 +1,217 @@
...
+    if (errorOrMaybeService.isLeft) {
+      return ResponseErrorQuery(
+        "Error trying to retrieve existing service",
+        errorOrMaybeService.left
+      );
+    }
+
+    const maybeService = errorOrMaybeService.right;
+    if (maybeService.isEmpty) {
+      return ResponseErrorValidation(
```
cloudify: this should be a 404 actually

lib/controllers/admin_services.ts
```diff
@@ -0,0 +1,217 @@
...
+    return ResponseSuccessJson(maybeUpdatedService.get);
+  };
+}
+
+export function CreateServiceHandler(
+  serviceModel: ServiceModel,
+  generateObjectId: ObjectIdGenerator
+): ICreateServiceHandler {
+  return async (_, serviceModelPayload) => {
+    if (serviceModelPayload.serviceId) {
```
cloudify: I think it makes sense to have two different types here (new and existing)

lib/controllers/messages.ts
```diff
@@ -211,7 +211,10 @@ export function CreateMessageHandler(
     if (auth.groups.has(UserGroup.ApiLimitedMessageWrite)) {
       // user is in limited message creation mode, check whether he's sending
       // the message to an authorized recipient
-      if (!userAttributes.authorizedRecipients.has(fiscalCode)) {
+      if (
+        !userAttributes.service.authorizedRecipients ||
+        userAttributes.service.authorizedRecipients.indexOf(fiscalCode) === -1
```
cloudify: uhm, why? `authorizedRecipients` is a `Set`

lib/models/service.ts
```diff
@@ -0,0 +1,165 @@
...
+export interface IService {
+  readonly serviceId: NonEmptyString;
+  // the name of the department within the service
+  readonly departmentName: NonEmptyString;
+  // the name of the service
+  readonly serviceName: NonEmptyString;
+  // the name of the organization
+  readonly organizationName: NonEmptyString;
+  // list of authorized fiscal codes
+  readonly authorizedRecipients?: ReadonlyArray<FiscalCode>;
```
cloudify: this should be a `ReadonlySet` (a recipient is supposed to appear once in the collection)

lib/utils/documentdb.ts
```diff
@@ -224,7 +224,11 @@ export function queryDocuments<T>(
   collectionUri: IDocumentDbCollectionUri,
   query: DocumentDb.DocumentQuery
 ): IResultIterator<T & DocumentDb.RetrievedDocument> {
-  const documentIterator = client.queryDocuments(collectionUri.uri, query);
+  const documentIterator = client.queryDocuments(collectionUri.uri, query, {
+    // enableCrossPartitionQuery is mandatory if you want to query for fields
+    // other than the partitionKey and Id
+    enableCrossPartitionQuery: true
```
cloudify: this has certainly performance implications, I would expose it as a parameter of `queryDocuments`, with a default value of `false`, letting the caller decide

lib/utils/documentdb_model_versioned.ts
```diff
@@ -94,9 +99,9 @@ export abstract class DocumentDbModelVersioned<
   public async update(
     objectId: string,
     partitionKey: string,
-    f: (current: T) => T
+    f: (current: T) => T | TR
```
cloudify: why `TR`? `TR` is `T` + storage specific attributes, the update functions shouldn't have any knowledge about them

lib/utils/middlewares/azure_api_auth.ts
```diff
@@ -31,10 +31,10 @@ export enum UserGroup {
...
   ApiProfileWrite = "ApiProfileWrite",
 
-  // organizations: read organization attributes
-  ApiOrganizationRead = "ApiOrganizationRead",
-  // organizations: create and update organizations
-  ApiOrganizationWrite = "ApiOrganizationWrite",
+  // services: read services attributes
+  ApiServiceRead = "ApiServiceRead",
+  // services: create and update services
+  ApiServiceWrite = "ApiServiceWrite",
```
cloudify: we must also rename the corresponding groups in the Azure API Management service

lib/utils/middlewares/path_param.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * A request middleware that validates the presence of a valid `param` parameter
+ * in the request. In case the parameter is missing or is not valid, the middleware
+ * returns an `IResponseErrorValidation`.
+ */
+export const PathParamMiddleware: <T>(
+  paramName: string,
+  // tslint:disable-next-line:no-any
+  testfn: (arg: any) => arg is T
```
cloudify: I think you just need a `(string) => boolean` here

lib/models/service.ts
```diff
@@ -0,0 +1,165 @@
...
+    // tslint:disable-next-line:no-object-mutation
+    this.collectionUri = collectionUrl;
+  }
+
+  /**
+   * Searches for one Service associated to the provided ID
+   *
+   * @param fiscalCode
+   */
+  public findByServiceId(
```
cloudify: Why don't we just keep `serviceId` == `subscriptionId` and avoid a having two IDs?

AdminApi/function.json
gunzip: yes, I haven't merged master branch yet

AdminApi/function.json
gunzip: ok for get rid of versioning, but the path "admin" is reserved for internal use (that was my first try)

lib/admin_api.ts
gunzip: why ?

lib/controllers/admin_services.ts
gunzip: ok

lib/controllers/admin_services.ts
gunzip: ok

lib/controllers/admin_services.ts
gunzip: ok

lib/controllers/admin_services.ts
gunzip: why new if updating ?

lib/controllers/messages.ts
gunzip: yes I've changed it because now it's retrieved from the document in cosmosdb (as a plain array). 
of course we can make a Set() from the array but it's probably more expensive than just call indexOf here.

lib/models/service.ts
gunzip: see comment above

lib/models/service.ts
gunzip: because I misundestood this comment:  "Unica modifica che farei è non usare subscriptionId per l'id, ma avere un campo azureSubscriptionId, così sleghiamo l'esistenza di Service dalla subscriptionId (che un domani possiamo eliminare completamente come attributo)." 

I will get rid of serviceId field

lib/utils/documentdb.ts
gunzip: ok, I'll revert this back setting subscriptionId = serviceId

lib/utils/middlewares/azure_api_auth.ts
gunzip: ok

lib/utils/middlewares/path_param.ts
gunzip: maybe it's clearer with the current signature

AdminApi/function.json
cloudify: ok

lib/utils/documentdb_model_versioned.ts
gunzip: because TR has an 'id'. when we update the document it's a retrieved-document (or you have to get rid of the id before updating it, which is unpractical)

lib/admin_api.ts
cloudify: it's a better design - because that's the minimum information needed by `getSecureExpressApp` function to be useful, your intent was to refactor the security configurations into a common function, not the instantiation of an express app - always go for the minimal dependency / responsibility

lib/controllers/admin_services.ts
cloudify: the fact that `New` is in the name is misleading, think about that type as`IServiceWithoutId` 😄 

the problem in using `IRetrievedService` here is that it extends `DocumentDb.RetrievedDocument` and `IVersionedModel`... these interfaces contain attributes that should not be a concern of the client

lib/controllers/messages.ts
cloudify: I suggest to keep a `Set`, since an `Array` is semantically wrong

lib/models/service.ts
cloudify: I think that was my comment 😄 but looking at the code now (and the issue with the multi partition query), I would not agree with it anymore

lib/utils/documentdb_model_versioned.ts
cloudify: in fact that's what happens, the ID get's stripped from the retrieved document by `toBaseType`

lib/utils/middlewares/path_param.ts
cloudify: why ask for an `any`, when you know you're going to pass a `string`? always go for the narrower type, avoid asking for an `any` at all costs

also, you don't need a type guard here, a function that simply returns a `boolean` is sufficient (that doesn't mean you can't still pass a type guard here) - the principle is to make the types and parameter as few and as more specific as possibile

lib/utils/documentdb_model_versioned.ts
gunzip: ok

lib/utils/express.ts
gunzip: ok

lib/controllers/adm/services.ts
```diff
@@ -0,0 +1,228 @@
...
+/**
+ * Wraps an UpdateService handler inside an Express request handler.
+ */
+export function UpdateService(
+  serviceModel: ServiceModel
+): express.RequestHandler {
+  const handler = UpdateServiceHandler(serviceModel);
+  const middlewaresWrap = withRequestMiddlewares(
+    AzureApiAuthMiddleware(new Set([UserGroup.ApiServiceWrite])),
+    RequiredParamMiddleware(params => {
```
cloudify: for clarity, pls instantiate this middleware outside `UpdateService`, i.e. `const requiredServiceIdMiddleware = ...`

lib/controllers/messages.ts
```diff
@@ -211,7 +211,10 @@ export function CreateMessageHandler(
     if (auth.groups.has(UserGroup.ApiLimitedMessageWrite)) {
       // user is in limited message creation mode, check whether he's sending
       // the message to an authorized recipient
-      if (!userAttributes.authorizedRecipients.has(fiscalCode)) {
+      if (
+        !userAttributes.service.authorizedRecipients ||
```
cloudify: when is `userAttributes.service.authorizedRecipients` not defined?

lib/models/service.ts
```diff
@@ -0,0 +1,166 @@
...
+  readonly organizationName: NonEmptyString;
+}
+
+/**
+ * Interface needed to interact with
+ * retrieved services from database
+ */
+export interface ISerializableService extends IBaseService {
+  // list of authorized fiscal codes
+  readonly authorizedRecipients?: ReadonlyArray<FiscalCode>;
```
cloudify: The problem with Sets that are not correctly serialized by `JSON.stringify` can be solved using [serializable Sets](https://www.npmjs.com/package/json-set-map) without having to define a separate `ISerializableService` type.

lib/utils/express.ts
```diff
@@ -0,0 +1,36 @@
+import * as express from "express";
+import * as helmet from "helmet";
+import * as csp from "helmet-csp";
+import * as referrerPolicy from "referrer-policy";
+
+export function secureExpressApp(app: express.Express): void {
```
cloudify: this needs a short comment explaining what it does

lib/controllers/adm/services.ts
gunzip: ok

lib/utils/express.ts
gunzip: ok

lib/models/service.ts
gunzip: this works if you use JSON.stringify() but we need the serialization part because all the DocumentDb methods work with plain json objects (ie. when you call a find() it returns a ISerializableService with an array and not a Set and that's the same with update() and create())

cloudify: the conversion from a DocumentDB document to our data type can be handled in the `toRetrieved` method defined in the model class

gunzip: what about the inverse conversion (to pass a value to update() and create()) ?

cloudify: you can pass the object `T`, the DocumentDB client relies on `JSON.stringify` to serialize the object, thus if `Set` gets correctly serialized you're good

gunzip: I'm trying to follow the path you've suggested but

cloudify: that's probably a bug, it should pass through `toRetrieved`

gunzip: what about this refactoring:

https://gist.github.com/gunzip/8718322cc7e7ab80161d6d1a16cb8b90

cloudify: LGTM

lib/utils/middlewares/azure_api_auth.ts
gunzip: done.

lib/controllers/adm/services.ts
```diff
@@ -70,13 +66,13 @@ type IUpdateServiceHandler = (
...
+export const ServicePayloadMiddleware: IRequestMiddleware<
   IResponseErrorValidation,
-  ISerializableService
+  IService
 > = request => {
   const body = request.body;
-  const servicePayload: ISerializableService = {
-    authorizedRecipients: body.authorized_recipients,
+  const servicePayload: IService = {
+    authorizedRecipients: toAuthorizedRecipientsSet(body.authorized_recipients),
```
cloudify: avoid leaking the data structure to the method, if tomorrow we're going to use a different data structure than a Set, we don't want to change the method name... `toAuthorizedRecipients` would be better

lib/controllers/adm/services.ts
```diff
@@ -116,6 +112,13 @@ export function UpdateServiceHandler(
   serviceModel: ServiceModel
 ): IUpdateServiceHandler {
   return async (_, serviceId, serviceModelPayload) => {
+    if (serviceModelPayload.serviceId !== serviceId) {
+      return ResponseErrorValidation(
+        "Error validating payload",
+        "Value of `service_id` in the request body must match " +
```
cloudify: intentional use of backticks? no markdown here...

lib/controllers/adm/services.ts
```diff
@@ -212,16 +207,17 @@ export function UpdateService(
   serviceModel: ServiceModel
 ): express.RequestHandler {
   const handler = UpdateServiceHandler(serviceModel);
+  const requiredserviceIdMiddleware = RequiredParamMiddleware(params => {
```
cloudify: `requiredServiceIdMiddleware`

lib/models/service.ts
```diff
@@ -54,53 +49,56 @@ export interface INewService
...
+ * 
+ * We need to handle Arrays as this method is called by database finders
+ * who retrieve a plain json object.
+ * 
+ * We need to handle Sets as this method is called on IService objects
+ * passed to create(IService) and update(IService) model methods.
+ * 
+ * @param authorizedRecipients 
+ */
+export function toAuthorizedRecipientsSet(
```
cloudify: This is a little ugly, the method accepts the union of two types but it implements two different logics for the two types (so what's the advantage of having a single method accepting an union type?)

Having two separate methods would make it explicit that the two sources are handled differently.

lib/models/service.ts
```diff
@@ -67,22 +67,20 @@ export interface IRetrievedService
...
+export function toAuthorizedRecipients(
   authorizedRecipients: ReadonlyArray<string> | ReadonlySet<string>
 ): ReadonlySet<FiscalCode> {
-  return authorizedRecipients && authorizedRecipients instanceof Array
-    ? new Set(authorizedRecipients.filter(isFiscalCode))
-    : authorizedRecipients instanceof Set
-      ? new Set(Array.from(authorizedRecipients).filter(isFiscalCode))
-      : new Set();
+  return new Set(
+    Array.from(authorizedRecipients ? authorizedRecipients : []).filter(
```
cloudify: the type of `authorizedRecipients` should reflect that it may be `undefined`: `ReadonlyArray<string> | ReadonlySet<string> | undefined`, unless you call `toAuthorizedRecipients` only when `authorizedRecipients` is defined

lib/models/service.ts
```diff
@@ -67,22 +67,20 @@ export interface IRetrievedService
...
+  );
 }
 
 function toRetrieved(result: DocumentDb.RetrievedDocument): IRetrievedService {
   return {
     ...result,
-    authorizedRecipients: toAuthorizedRecipientsSet(
-      result.authorizedRecipients
-    ),
+    authorizedRecipients: toAuthorizedRecipients(result.authorizedRecipients),
```
cloudify: (see comment above) this may become

```typescript
authorizedRecipients: result.authorizedRecipients ? toAuthorizedRecipients(result.authorizedRecipients) : new Set() 
```

lib/models/service.ts
gunzip: I've added undefined to accepted types

PR #88, opened 2017-10-27
fix: make gulp test task to not freeze on win32

gunzip:


gulpfile.js
```diff
@@ -96,9 +97,11 @@ gulp.task("yarn:build", () => {
 /**
  * Run the test task
  */
-gulp.task("yarn:test", () => {
+gulp.task("jest:test", () => {
```
cloudify: this is duplicating the test logic (`yarn test` will also run `ts-lint`), if you want to move the `test` logic here, you should remove the `test` script from `package.json` to avoid confusion (and also add linting)

gulpfile.js
```diff
@@ -95,9 +95,18 @@ gulp.task("yarn:build", () => {
...
+  return gulp.src(TYPESCRIPT_SOURCE_DIR)
+    .pipe(exec(`yarn run lint`))
+    .pipe(exec.reporter());
+});
+
+/**
  * Run the test task
  */
-gulp.task("jest:test", () => {
+gulp.task("jest:test", ["yarn:lint"], () => {
```
cloudify: I would just name it `test`, or at least have an alias just named `test`.

Also worth updating [this paragraph](https://github.com/teamdigitale/digital-citizenship-functions#unit-tests) with instructions on how to run the tests.

gunzip: I've refactored the script a little to make room for other kinds of test (other than unit) in future.

PR #91, opened 2017-11-09
Fix upsert profile

gunzip:
fixes a nasty exception escapes from middleware when request.body is empty or email is undefined.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "digital-citizenship-functions",
-  "version": "0.10.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "digital-citizenship-functions",
-  "version": "0.10.0-SNAPSHOT",
+  "version": "0.11.0-SNAPSHOT",
```
cloudify: why the version bump?

PR #92, opened 2017-11-10
chore: added get service endpoint

gunzip:


lib/controllers/adm/services.ts
```diff
@@ -168,6 +172,15 @@ export function UpdateServiceHandler(
   };
 }
 
+const requiredServiceIdMiddleware = RequiredParamMiddleware(params => {
```
cloudify: + comment

lib/controllers/adm/services.ts
```diff
@@ -184,6 +197,44 @@ export function CreateServiceHandler(
...
+      return ResponseErrorQuery(
+        "Error while retrieving the service",
+        errorOrMaybeService.left
+      );
+    }
+  };
+}
+
+/**
+ * Wraps a CreateService handler inside an Express request handler.
```
cloudify: `CreateService` -> `GetService`

PR #93, opened 2017-11-11
chore: added examples to api spec

gunzip:
- added examples to swagger specs
- added upsert profile endpoint to specs
- renamed sender_organization_id to sender_service_id

### DIFF
--- a/api/public_api_v1.yaml
+++ b/api/public_api_v1.yaml
@@ -38,7 +38,7 @@ paths:
                   subject: aliquip sint nulla in estinut
                   markdown: Excepteur estinut inincididunt laboris culpa officiaeuexercitation
                     quisLorem in et est
-                sender_organization_id: '1'
+                sender_service_id: '1'
               notification:
                 email: QUEUED
         '404':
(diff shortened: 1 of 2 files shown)


lib/api/public_api_v1.ts
```diff
@@ -35,7 +35,22 @@ export const specs = {
           "200": {
             description: "Message found.",
             schema: { $ref: "#/definitions/MessageResponse" },
-            examples: {}
+            examples: {
+              "application/json": {
+                message: {
+                  id: "1",
```
cloudify: I would give an ULID as example to avoid interpreting this as an integer :)

api/public_api_v1.yaml
```diff
@@ -38,7 +38,7 @@ paths:
                   subject: aliquip sint nulla in estinut
                   markdown: Excepteur estinut inincididunt laboris culpa officiaeuexercitation
                     quisLorem in et est
-                sender_organization_id: '1'
+                sender_service_id: '1'
```
cloudify: better use an alphanumeric value

PR #100, opened 2017-11-17
[#150777493] set logging level from env

cloudify:
https://www.pivotaltracker.com/story/show/150777493

### DIFF
--- a/lib/admin_api.ts
+++ b/lib/admin_api.ts
@@ -28,6 +28,9 @@ import { GetDebug } from "./controllers/debug";
 import * as express from "express";
 import { secureExpressApp } from "./utils/express";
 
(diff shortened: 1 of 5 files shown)


lib/admin_api.ts
```diff
@@ -66,7 +66,9 @@ const azureFunctionHandler = createAzureFunctionHandler(app);
 
 // Binds the express app to an Azure Function handler
 export function index(context: IContext<{}>): void {
-  configureAzureContextTransport(context, winston, "debug");
+  const isProduction = process.env.NODE_ENV === "production";
```
gunzip: I would put this line at the module level, just near the other initialization from env vars

PR #103, opened 2017-11-20
Removes ts-option in favor of fp-ts

cloudify:


lib/controllers/messages.ts
```diff
@@ -442,8 +443,8 @@ export function GetMessageHandler(
 
     const messageStatus: MessageResponse = {
       message: retrievedMessageToPublic(retrievedMessage),
-      notification: maybeNotificationStatus.isDefined
-        ? maybeNotificationStatus.get
+      notification: maybeNotificationStatus.isSome()
```
gunzip: maybeNotificationStatus.toUndefined() without the isSome() check should suffice

lib/utils/env.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Helper function that validates an environment variable and return its value
+ * if it's a `NonEmptyString`.
+ * Throws an Error otherwise.
+ */
+export function getRequiredStringEnv(k: string): NonEmptyString {
+  const maybeValue = toNonEmptyString(process.env[k]);
+
+  if (isNone(maybeValue)) {
+    throw new Error(`${k} must be defined and non-empty`);
```
gunzip: maybe a call to winston log() here would be useful as well ?

lib/utils/env.ts
```diff
@@ -0,0 +1,17 @@
+import { isNone } from "fp-ts/lib/Option";
+import { NonEmptyString, toNonEmptyString } from "./strings";
+
+/**
+ * Helper function that validates an environment variable and return its value
+ * if it's a `NonEmptyString`.
+ * Throws an Error otherwise.
+ */
+export function getRequiredStringEnv(k: string): NonEmptyString {
```
gunzip: better name getRequiredEnvString ?

lib/utils/middlewares/azure_api_auth.ts
```diff
@@ -81,8 +81,8 @@ function getGroupsFromHeader(groupsHeader: string): Set<UserGroup> {
     groupsHeader
       .split(",")
       .map(v => toUserGroup(v))
-      .filter(g => g.isDefined)
-      .map(g => g.get)
+      .filter(g => isSome(g))
+      .map(g => (g as Some<UserGroup>).value)
```
gunzip: maybe typescript can infer the type using objects methods ie.  g.isSome() instead of isSome(g) ?

cloudify: unfortunately it does not

lib/utils/env.ts
cloudify: the exception should go into the logs already, doesn't it?

PR #104, opened 2017-11-24
chore: [#153005275] added admin API specs to OpenAPI

gunzip:
This PR enables the API managament automatic integration with Admin API (ie. to manage services) (actually the admin API are created manually in the APIm interface).

- added admin api specs
- changed openapi specs path (added /specs prefix)
- modified nunjuck template to generate models with arrays


api/generate_models.ts
```diff
@@ -33,6 +28,12 @@ async function generateModelsFromApi(
     })
   );
 
+  const definitions = api.definitions;
+  if (!definitions) {
+    console.log("No definitions found");
```
cloudify: I would say `No definitions found, skipping generation of model code` or something like that

api/generate_models.ts
```diff
@@ -68,3 +69,8 @@ generateModelsFromApi(env, "public_api_v1", "lib/api").then(
   () => console.log("done"),
   err => console.log(`Error: ${err}`)
 );
+
+generateModelsFromApi(env, "admin_api", "lib/api").then(
```
cloudify: it looks like this method responsibility is not just generating models anymore, it's worth changing its name to something else, like `exportApi` or `generateApi`

api/admin_api.yaml
```diff
@@ -0,0 +1,298 @@
...
+    - organizationName
+    - departmentName
+    - authorizedCIDRs
+    - authorizedRecipients
+  AdmCidr:
+    type: string
+    title: AdmCidr
+    description: Describes a single IP or a range of IPs.
+    pattern: "([0-9]{1,3}\\.){3}[0-9]{1,3}(\\/([0-9]|[1-2][0-9]|3[0-2]))?"
+  AdmFiscalCode:
```
cloudify: why we need an `AdmFiscalCode`? we already have a definition for `FiscalCode`

api/admin_api.yaml
```diff
@@ -0,0 +1,298 @@
...
+      id:
+        type: string
+    required:
+    - serviceId
+    - serviceName
+    - organizationName
+    - departmentName
+    - authorizedCIDRs
+    - authorizedRecipients
+  AdmCidr:
```
cloudify: can we call it `CIDR` ?

api/admin_api.yaml
gunzip: I opted to keep the API spec "self contained", othwerwise we got:

    $ref : "public_api_v1.yaml#/definitions/FiscalCode"

which introduces some non-trivial problems when generating the models from the specs
(even calling Swagger.bundle() instead of Swagger.parse() does not solve the issue).

That's the reason why I opted for using a prefix/namespace ("Adm") to the models definitions, in this way they does not conflict with the other ones.

I suggest to keep thie behavoir and open an issue if we really need to share definitions between specs files (which is a broader problem than the aim of this PR)

api/admin_api.yaml
gunzip: see comment below

api/generate_models.ts
gunzip: ok

api/generate_models.ts
gunzip: ok

api/admin_api.yaml
gunzip: I've updated the specs extracting the definitions in another YAML file. The drawback is that the whole definition file is bundled into the two specs (even the unused definitions). Aside the bigger swagger.json size it does not impact any other aspect.

PR #105, opened 2017-11-25
chore: clarified ip checking middleware

gunzip:
added some docs in comment and clarified error in case the ip cannot be extracted from request

### DIFF
--- a/lib/utils/middlewares/client_ip_middleware.ts
+++ b/lib/utils/middlewares/client_ip_middleware.ts
@@ -7,13 +7,20 @@ import { IRequestMiddleware } from "../request_middleware";
 
 export type ClientIp = Option<IPString>;
 
+import * as winston from "winston";
+
 /**
- * A middleware that extracts the client IP from the request
+ * A middleware that extracts the client IP from the request.
+ *
+ * The IP is extracted from the "x-forwarded-for" HTTP header;
+ * if x-forwarded-for header is empty, the IP will be null 
+ * and the middleware will return None.
  */
 export const ClientIpMiddleware: IRequestMiddleware<
   never,
   ClientIp
 > = request => {
   const clientIp = requestIp.getClientIp(request);
+  winston.debug(`Handling request for client ip|${clientIp}`);
   return Promise.resolve(right<never, ClientIp>(toIPString(clientIp)));
 };
--- a/lib/utils/source_ip_check.ts
+++ b/lib/utils/source_ip_check.ts
@@ -1,4 +1,4 @@
-import { isSome } from "fp-ts/lib/Option";
+import { isNone } from "fp-ts/lib/Option";
 import { IAzureUserAttributes } from "./middlewares/azure_user_attributes";
 /*
  * A request wrapper that checks whether the source IP is contained in the
@@ -12,7 +12,9 @@ import { Tuple2 } from "./tuples";
 
 import {
   IResponseErrorForbiddenNotAuthorized,
-  ResponseErrorForbiddenNotAuthorized
+  IResponseErrorInternal,
+  ResponseErrorForbiddenNotAuthorized,
+  ResponseErrorInternal
 } from "./response";
 
 /**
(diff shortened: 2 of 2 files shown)


lib/utils/middlewares/client_ip_middleware.ts
```diff
@@ -7,13 +7,20 @@ import { IRequestMiddleware } from "../request_middleware";
 
 export type ClientIp = Option<IPString>;
 
+import * as winston from "winston";
+
 /**
- * A middleware that extracts the client IP from the request
+ * A middleware that extracts the client IP from the request.
+ *
+ * The IP is extracted from the "x-forwarded-for" HTTP header;
```
cloudify: That's not correct, [here's how it works](https://www.npmjs.com/package/request-ip#how-it-works) - if you want you can just link to that page

lib/utils/middlewares/client_ip_middleware.ts
```diff
@@ -7,13 +7,20 @@ import { IRequestMiddleware } from "../request_middleware";
...
+ * if x-forwarded-for header is empty, the IP will be null 
+ * and the middleware will return None.
  */
 export const ClientIpMiddleware: IRequestMiddleware<
   never,
   ClientIp
 > = request => {
   const clientIp = requestIp.getClientIp(request);
+  winston.debug(`Handling request for client ip|${clientIp}`);
```
cloudify: `ip` -> `IP`

lib/utils/source_ip_check.ts
```diff
@@ -116,25 +118,36 @@ export function checkSourceIpForHandler<P1, P2, P3, P4, P5, P6, O>(
...
-) => Promise<O | IResponseErrorForbiddenNotAuthorized> {
+) => Promise<
+  O | IResponseErrorForbiddenNotAuthorized | IResponseErrorInternal
+> {
   return (p1: P1, p2: P2, p3: P3, p4: P4, p5: P5, p6: P6) => {
-    // extract the x-forwarded-for header and the allowed cidrs from the params
-    const x = extractor(p1, p2, p3, p4, p5, p6);
-    const maybeClientIp = x.e1;
-    const cidrs = x.e2;
+    return new Promise(resolve => {
```
cloudify: why you need this? there's no callback involved

lib/utils/source_ip_check.ts
```diff
@@ -116,25 +118,36 @@ export function checkSourceIpForHandler<P1, P2, P3, P4, P5, P6, O>(
...
-      // forward request to handler
-      return handler(p1, p2, p3, p4, p5, p6);
-    } else {
-      // respond with Not Authorized
-      return Promise.resolve(ResponseErrorForbiddenNotAuthorized);
-    }
+      if (isNone(maybeClientIp)) {
+        return resolve(
+          ResponseErrorInternal(
+            "IP address cannot be extracted from the request, check x-forwarder-for HTTP header"
```
cloudify: this error will reach the client, but the client doesn't know anything about `x-forwarded-for`...

cloudify: also this step should not run if `cidrs.size === 0`

lib/utils/source_ip_check.ts
gunzip: the middleware returns a promise and wrapping the whole code in a promise avoid the exceptions to "escape" the context and not get caught

lib/utils/middlewares/client_ip_middleware.ts
gunzip: ok, but that's how it works in our context atm (using azure api management). 
the point of this PR is that I got a generic "403 forbidden" once (because I've called the functions code directly bypassing the api gateway) and took it 1hr to debug why :)

lib/utils/source_ip_check.ts
gunzip: ok

lib/utils/middlewares/client_ip_middleware.ts
gunzip: ok

lib/utils/middlewares/client_ip_middleware.ts
cloudify: yes this is how the system uses this function, but it's not how this function works :)

so perhaps you should link to the `request-ip` documentation from here, and add the comment about `x-forwarded-for` to the README

lib/utils/source_ip_check.ts
gunzip: just notice that the actual behavior is to fail when the IP is not found (even if cidr.size == 0)

https://github.com/teamdigitale/digital-citizenship-functions/blob/master/lib/utils/source_ip_check.ts#L128

lib/utils/source_ip_check.ts
```diff
@@ -116,25 +118,36 @@ export function checkSourceIpForHandler<P1, P2, P3, P4, P5, P6, O>(
...
-    }
+      if (isNone(maybeClientIp)) {
+        return resolve(
+          ResponseErrorInternal(
+            "IP address cannot be extracted from the request"
+          )
+        );
+      }
+
+      if (
```
gunzip: I've kept the actual behavior (fail when no IP is found), tell me if you want to change it 
(pass the check if authorized_cidrs == 0 even when IP is null)

lib/utils/middlewares/client_ip_middleware.ts
gunzip: done

lib/utils/source_ip_check.ts
cloudify: obviously I'm not convinced about either option :D what's your opinion?

cloudify: makes sense but I would link that behavior only to a non-production environment (i.e.  `NODE_ENV !== "production"`)

PR #106, opened 2017-11-25
chore: added admin api specs

gunzip:


api/definitions.yaml
```diff
@@ -0,0 +1,260 @@
+ProblemJson:
```
ioggstream: @gunzip  why don't

 - just C&P https://zalando.github.io/problem/schema.yaml so that we could just diff it?
 - still define HttpStatus for other purposes

PR #107, opened 2017-11-27
chore: use yarn instead of npm

gunzip:
using npm results in failed deployment:

`Module '"D:/home/site/wwwroot/node_modules/ulid/index"' resolves to a non-module entity and cannot be imported using this construct.`

with yarn everything works. 

accidentally, it's fixed in production atm as I once ran yarn install from kudu console.


cloudify:
mmh wait, `node_modules` shouldn't exist in production, funcpack creates a single js file with all dependencies in it

PR #110, opened 2017-12-01
[#151318413] uses ts-io for type validation and error messages

cloudify:
Adds type validation of payloads and human readable error messages


lib/utils/types.ts
```diff
@@ -1,3 +1,11 @@
+import * as t from "io-ts";
+
+export type Tagged<T, S, A> = t.Type<S, A & T>;
+
+export const tag = <T>() => <S, A>(type: t.Type<S, A>): Tagged<T, S, A> =>
```
gunzip: it looks using HTK is the idiomatic way to define tagged types with io-ts:

https://github.com/gcanti/fp-ts/blob/master/docs/HKT.md

cloudify: example?

cloudify: btw this approach [was suggested](https://github.com/gcanti/io-ts/issues/97) by the author of io-ts

gunzip: ok

api/templates/model.ts.njk
```diff
@@ -24,14 +24,14 @@ import { fromNullable, Option } from "fp-ts/lib/Option";
 
 export interface {{ definitionName }} {
 {% for propName, prop in definition.properties -%}
```
gunzip: actually, the logic to handle object properties is non-recursive. this means two thing:

1. it only works for first level object properties (so you must create a dedicated object definition for every nested property)
2. you have to duplicate all the validation logic for object properties *and* for single properties

consider to handle all this logic using nunjucks macros: 
https://mozilla.github.io/nunjucks/templating.html#macro 
(one for every property type and a recursive macro for objects)

lib/utils/types.ts
gunzip: Is there a difference with using newtype-js ?

lib/utils/response.ts
```diff
@@ -192,6 +193,17 @@ export function ResponseErrorValidation(
   };
 }
 
+/**
+ * Returns a response describing a validation error.
+ */
+export function ResponseErrorValidationFromValidation<T>(
```
gunzip: better naming: ResponseErrorValidationPayload ?

cloudify: I was actually thinking of removing the current `ResponseErrorValidation` and name this one as `ResponseErrorValidation` instead, I have to check if we still use the current one once I've replace all payload validations

lib/utils/types.ts
cloudify: conceptually should be similar, newtype is based on `monocle-ts` that is very opinionated

gcanti: It depends on the expected behavior, a tagged (or branded) type `A & { kind: 'MyTag' }` can be used where a value of type `A` is required whereas newtypes are (kind of) opaque and must be projected

```ts
type TaggedInteger = number & { kind: 'Integer' }
type NewtypedInteger = { kind: 'Integer' }

declare function double(n: number): number
declare const x: TaggedInteger
declare const y: NewtypedInteger
double(x) // ok
double(y) // nope

const prj = (x: NewtypedInteger): number => x as any
double(prj(y)) // ok, returns a number though
```

Based on the example in the linked issue

```ts
function f(i: Integer): void {
  console.log(i);
}
```

I wasn't sure of the requirements. 

So **if** you want to fake newtypes, you can amend the `tag` function

```ts
export const tag = <T>() => <S>(type: t.Type<S, any>): t.Type<S, T> => type as any
```

However newtypes are pretty awkward to operate on, you have to wrap/unwrap them all the times. `newtype-ts` leverages an `Iso` to avoid this boilerplate (and gain some other useful features like `modify`). There's no need to add `monocle-ts` as a dependency if you are just interested in `Iso`, you can define your own

```ts
class Iso<S, A> {
  constructor(readonly get: (s: S) => A, readonly reverseGet: (a: A) => S) {}
  modify(f: (a: A) => A): (s: S) => S {
    return s => this.reverseGet(f(this.get(s)))
  }
}

const unsafeCoerce = <A, B>(a: A): B => a as any
// returns an Iso for the newtype T with carrier A
const iso = <T, A>(): Iso<T, A> => new Iso<T, A>(unsafeCoerce, unsafeCoerce)

// usage
const newtypedIntegerIso = iso<NewtypedInteger, number>()

const z1: number = double(newtypedIntegerIso.get(y))
// ...or you can lift `double` using `modify`
const z2: NewtypedInteger = newtypedIntegerIso.modify(double)(y)
```

PR #111, opened 2017-12-01
refactor: make service optional

gunzip:
This PR makes Service(s) optional. Why ? Because otherwise we have a chicken-egg problem: actually the APIs are callable only when there's a service tied to the user's subscription, so you cannot create a service,  using the API, if there's no service already setup for the caller account. This causes a "deadlock" in the 
https://github.com/teamdigitale/digital-citizenship-onboarding web application that needs to create an administrative account to manage API users programmatically.

### DIFF
--- a/lib/controllers/__tests__/messages.test.ts
+++ b/lib/controllers/__tests__/messages.test.ts
@@ -65,14 +65,14 @@ const aMessageBodyMarkdown = _getO(toMessageBodyMarkdown("test".repeat(80)));
 const someUserAttributes: IAzureUserAttributes = {
   email: anEmail,
   kind: "IAzureUserAttributes",
-  service: {
+  service: some({
     authorizedCIDRs: toAuthorizedCIDRs([]),
     authorizedRecipients: new Set([]),
     departmentName: _getO(toNonEmptyString("IT")),
     organizationName: _getO(toNonEmptyString("AgID")),
     serviceId: _getO(toNonEmptyString("test")),
     serviceName: _getO(toNonEmptyString("Test"))
-  }
+  })
 };
 
 const aUserAuthenticationDeveloper: IAzureApiAuthorization = {
@@ -252,10 +252,10 @@ describe("CreateMessageHandler", () => {
       log: jest.fn()
     };
 
-    const anAuthorizedService = {
-      ...someUserAttributes.service,
+    const anAuthorizedService = someUserAttributes.service.map(service => ({
+      ...service,
       authorizedRecipients: new Set([aFiscalCode])
-    };
+    }));
 
     const someAuthorizedUserAttributes = {
       ...someUserAttributes,
--- a/lib/controllers/messages.ts
+++ b/lib/controllers/messages.ts
@@ -212,7 +212,16 @@ export function CreateMessageHandler(
     messagePayload
   ) => {
     // extract the user service
-    const userService = userAttributes.service;
+    const maybeUserService = userAttributes.service;
+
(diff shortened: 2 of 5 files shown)


cloudify:
mmh I'm not sure this is really an issue that must leak into the data model, the way you usually deal with this is that during the system "bootstrap" you init the database with known users, i.e. we can create an admin service by creating a record in the database.

most systems (i.e. databases) do that when you install them

gunzip:
ok I'll try. just notice that this will couple the code to the system bootsrap repository (digital-citizenhip) to the model layer of digital-citizenship-functions if we want to reuse the CRUD methods there.

cloudify:
you can split the two, have the generic infrastructure setup in `digital-citizenship` and have the API specific setup code in `digital-citizenship-functions` (i.e. a gulp task that bootstraps the database)

gunzip:
Can you please elaborate more ?

Let's suppose digital-citizenship repository brings the generic setup tasks:
```
git clone digital-citizenship repository
npm deploy-resources
```

This put cosmosdb, functions and the digital-citizenship-onboarding web application in place.

Now we need to setup the API key of a newly created user into the digital-citizenship-onboarding web application settings:
```
git clone digital-citizenship-functions
gulp create-api-user
```

suppose this output the API key of the user to the system terminal console.

```
cd ../digital-citizenship
set API_KEY=<the new API key>
npm continue-to-deploy
```

I think this workflow is suboptimal, it would be far better to just run the first step to put everything in place (without the need to locally clone the whole functions repository to just create the API user) .

cloudify:
what this `API_KEY` is needed for?

cloudify:
Can `npm deploy-resources` do the following?

1. create infrastructure resources (API Management, CosmosDB, etc...)
2. create an admin subscription
3. create an `admin` service associated to `2`
4. output the API key of `2`

then you can run the `digital-citizenship-onboarding` setup script with the provided API key

gunzip:
`npm deploy-resources` (in digital-citizenship repository) cannot create an admin service if not using digital-citizenship-functions model layer (that's exactly the problem :) 

I've created a script that does exactly that, but it's not optimal because it means you have to include as dependency the whole digital-citizienship-functions source code (in digital-citizenship) and compile it
just to do take that step:

https://github.com/gunzip/digital-citizenship/blob/apim-conf-db/lib/service.ts#L78

gunzip:
the API-KEY is needed in digital-citizenship-onboarding to manager API management users subscriptions programmatically (to create users' services and send welcome message)

cloudify:
To make things super clean we have to externalize the data model to its own package (and have it as dependency to both `digital-citizenship` and `digital-citizenship-functions`)...

BUT I think we can live an hardcoded JSON in `digital-citizenship` for now.

The other reasonable option: make the seed process manual, provide an example JSON in the README and let the operator do that through the Azure CosmosDB console.

It's ok if we just automate 99% of the things :)

gunzip:
that's not that easy because the service model is versioned "somehow" and this is only know to the dc model layer. both the solutions require this deep knowledge =)

cloudify:
yes but here we're talking about creating one single `Service` object in one single empty collection, you don't need to know how versioning works to do that, you just need a JSON template.

```json
{
  "service_id": subscriptionId,
  "service_name": "admin",
  "organization_name": "admin",
  "department_name": "admin",
  "authorized_cidrs": [],
  "authorized_recipients": [],
  "version": 1,
  "id": `${subscriptionId}-1`
}
```

PR #114, opened 2017-12-03
fix: do not use functions storage account for storing message content

gunzip:
as for queues, avoid using functions storage account to store message content


lib/created_message_queue_handler.ts
```diff
@@ -73,7 +73,7 @@ const notificationsCollectionUrl = documentDbUtils.getCollectionUri(
   "notifications"
 );
 
-const storageConnectionString = getRequiredStringEnv("AzureWebJobsStorage");
+const storageConnectionString = getRequiredStringEnv("QueueStorageConnection");
```
cloudify: No documentation to update?

gunzip: there are no references to the different blob storages actually

cloudify: I mean, in the setup instructions

gunzip: same there (no mention of the two storages as its handled by the programmatic tasks)

gunzip: can we merge this one ?

PR #115, opened 2017-12-04
[#151318413] validations

cloudify:


lib/utils/middlewares/fiscalcode.ts
```diff
@@ -13,19 +16,11 @@ import { IResponseErrorValidation, ResponseErrorValidation } from "../response";
...
-      right<IResponseErrorValidation, FiscalCode>(fiscalCode)
-    );
-  } else {
-    const validationErrorResponse = ResponseErrorValidation(
-      "Bad request",
-      `The fiscal code [${fiscalCode}] is not valid.`
-    );
-    return Promise.resolve(
-      left<IResponseErrorValidation, FiscalCode>(validationErrorResponse)
+> = request =>
```
gunzip: can it be RequiredParamMiddleware("fiscalcode", FiscalCode); ?

cloudify: actually yes! good catch

lib/utils/types.ts
```diff
@@ -26,6 +32,23 @@ export const createEnumType = <E>(e: {}, name: string): t.Type<any, E> => {
...
+export const readonlySetType = <S, E>(
+  o: t.Type<S, E>,
+  name: string
+): t.Type<any, ReadonlySet<E>> => {
+  const arrayType = t.readonlyArray(o, name);
+  return new t.Type<any, ReadonlySet<E>>(
+    name,
+    (v): v is ReadonlySet<E> => v instanceof Set && arrayType.is(Array.from(v)),
+    (s, c) =>
+      arrayType.is(s) ? t.success(new Set(Array.from(s))) : t.failure(s, c),
```
gunzip: ensure to 
`import { Set } from "json-set-map";` at the beginning of the file
otherwise you get unserializable Set()

lib/utils/middlewares/fiscalcode.ts
cloudify: done

lib/__tests__/created_message_queue_handler.test.ts
```diff
@@ -51,8 +52,8 @@ const aCorrectFiscalCode = _getO(
 );
 const aWrongFiscalCode = "FRLFRC74E04B157" as FiscalCode;
 const anEmail = _getO(t.validate("x@example.com", EmailString).toOption());
-const anEmailNotification: INotificationChannelEmail = {
-  addressSource: NotificationAddressSource.PROFILE_ADDRESS,
+const anEmailNotification: NotificationChannelEmail = {
+  addressSource: NotificationAddressSourceEnum.PROFILE_ADDRESS,
```
gunzip: better naming: is the "Enum" suffix mandatory ?

lib/models/message.ts
```diff
@@ -233,31 +167,25 @@ function toBaseType(o: IRetrievedMessage): IMessage {
...
-      kind: "IRetrievedMessageWithoutContent"
-    } as IRetrievedMessageWithoutContent;
+function toRetrieved(result: DocumentDb.RetrievedDocument): RetrievedMessage {
+  if (RetrievedMessageWithContent.is(result)) {
+    return result;
+  }
+  if (RetrievedMessageWithoutContent.is(result)) {
+    return result;
   }
+  throw new Error(
```
gunzip: throw ? cant' we return an option or either as stated in the old comments ?

lib/__tests__/created_message_queue_handler.test.ts
cloudify: yeah kind of, unfortunately TypeScript cannot export both a type and an `enum` with the same name (see [here](https://github.com/teamdigitale/digital-citizenship-functions/blob/master/lib/api/definitions/NotificationChannelStatus.ts))

lib/models/message.ts
cloudify: this is actually supposed to never happen, as you can see the old code was just casting the object, I decided to avoid the case and make the validation explicit

yes we could have `toRetrieved` return an `Option` but I'm not sure the added complexity makes it worth it - anyway, this is outside of the scope of this PR

PR #116, opened 2017-12-04
fix: do not return error on 404 not found in findLastVersionByModelId

gunzip:
[#153367500]
the unversioned find() method does not return an error when the document is not found.
backported the same behavior in findLastVersionByModelId()


cloudify:
Why this semantic change?

PR #119, opened 2017-12-15
[#151203322]: feat: added default values from openapi specs

gunzip:
calling t.validate() sets default values

### DIFF
--- a/api/templates/macros.njk
+++ b/api/templates/macros.njk
@@ -1,14 +1,14 @@
-{% macro exportConst(defaultValue, type, typedef) %}
+{% macro exportConst(defaultValue, type, typedef, quote = false) %}
   {% if defaultValue !== undefined %}
     {% set typeName = type + "X" %}
-    {% set q = '"' if type == "string" else "" %}
+    {% set q = '"' if quote else "" %}
     import * as t from "io-ts";
 
     import { withDefault } from "../../utils/default";
 
     const {{ typeName }} = {{ typedef | safe }};
 
-    const defaultValue = t.validate({{ q }}{{ defaultValue }}{{ q }}, {{ typeName }}).fold(_ => {
+    const defaultValue = t.validate({{ q | safe }}{{ defaultValue }}{{ q | safe }}, {{ typeName }}).fold(_ => {
       throw new Error("Invalid default value for {{ type }}");
     }, t.identity);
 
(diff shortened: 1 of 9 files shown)


lib/utils/default.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+// tslint:disable:no-any
+export function withDefault<T extends t.Any>(
+  type: T,
+  defaultValue: t.TypeOf<T>
+): t.Type<any, t.TypeOf<T>> {
+  return new t.Type(
+    type.name,
+    (v: any): v is T => type.is(v),
+    (v: any, c: any) => type.validate(v !== undefined ? v : defaultValue, c),
```
cloudify: `v !== undefined && v !== null` ?

lib/api/definitions/TimeToLive.ts
```diff
@@ -15,4 +17,7 @@ import { WithinRangeNumber } from "../../utils/numbers";
 
 export type TimeToLive = WithinRangeNumber<3600, 31536000>;
 
-export const TimeToLive = WithinRangeNumber(3600, 31536000);
+export const TimeToLive = withDefault(
+  WithinRangeNumber(3600, 31536000),
+  3600 as any
```
cloudify: this `as any` bothers me!

cloudify: you can replace `any` with `TimeToLive`, makes more sense and avoids using `any`

cloudify:
There's a little consistency problem, since default values get casted to a validated type, one could (by mistake) set a non valid value as default - to avoid this, the model generation script could validate the default values and trigger an error if the default value doesn't pass the type validation

lib/utils/default.ts
gunzip: probably it's better to not overwrite a property explicitely set to null, ie:

```
{
  foobar: null
}
```

gunzip: probably it's better to not overwrite any property explicitely set to null

cloudify: not sure about that, everywhere in the code there's no type for which `null` is a valid value, meaning that if `v === null`, validation fails

gunzip:
I've added a module defined variable `const defaultValue = ...` 
so the program exists with failure when the module is required (ie. during tests)

api/templates/macros.njk
```diff
@@ -0,0 +1,20 @@
+{% macro exportConst(defaultValue, type, typedef, quote = false) %}
+  {% if defaultValue !== undefined %}
+    {% set typeName = type + "X" %}
```
cloudify: perhaps a better naming here would be nice, like `Base` instead of `X`

gunzip:
done

PR #122, opened 2017-12-16
feat: #153291262 - save message storage preference

gunzip:


api/definitions.yaml
```diff
@@ -153,6 +153,11 @@ ExtendedProfile:
       $ref: "#/EmailAddress"
     preferred_languages:
       $ref: "#/PreferredLanguages"
+    is_storage_of_message_content_enabled:
```
cloudify: I would use a shorter name, like `inbox_enabled`, "whether the user enabled the inbox feature, for details on how the inbox feature works see..."

lib/api/admin_api.ts
```diff
@@ -417,6 +417,12 @@ export const specs = {
       properties: {
         email: { $ref: "#/definitions/EmailAddress" },
         preferred_languages: { $ref: "#/definitions/PreferredLanguages" },
+        is_storage_of_message_content_enabled: {
```
cloudify: `$ref` here too?

lib/api/public_api_v1.ts
```diff
@@ -380,6 +380,12 @@ export const specs = {
       properties: {
         email: { $ref: "#/definitions/EmailAddress" },
         preferred_languages: { $ref: "#/definitions/PreferredLanguages" },
+        is_storage_of_message_content_enabled: {
```
cloudify: `$ref` to definitions?

lib/controllers/profiles.ts
```diff
@@ -59,7 +59,9 @@ import { ServiceModel } from "../models/service";
 function toExtendedProfile(profile: RetrievedProfile): ExtendedProfile {
   return {
     email: profile.email,
-    version: profile.version
+    version: profile.version,
+    is_storage_of_message_content_enabled:
+      profile.isStorageOfMessageContentEnabled
```
cloudify: again I would give it a feature name, like `inbox` -> `isInboxEnabled`

"Inbox" may not be only about storing messages, the feature may provide other functionalities (e.g. search?), so it makes sense to use a name that is decoupled from a specific functionality

cloudify:
I'd like to merge PR #21 before this one

lib/api/admin_api.ts
gunzip: About using $ref or embedding the definition, actually there's no clear rule. Sometimes we use $ref(s) some others we embed the definitions (ie. CreatedMessage -> sender_service_id).

cloudify: when in doubt I would prefer avoiding duplication of definitions as they may go out of sync

cloudify:
linting errors, likely due to PR #121

PR #124, opened 2017-12-18
feat(#153291522): get message content from storage

gunzip:


lib/models/message.ts
```diff
@@ -301,4 +306,65 @@ export class MessageModel extends DocumentDbModel<
...
+      return left<Error, Option<MessageContent>>(
+        maybeContentAsTextOrError.value
+      );
+    }
+
+    // media exists but the content is empty
+    const maybeContentAsText = maybeContentAsTextOrError.value;
+    if (isNone(maybeContentAsText)) {
+      return left<Error, Option<MessageContent>>(
+        new Error("Cannot get stored message content from attachment")
```
cloudify: this is actually "message attachment is empty" ?

lib/models/message.ts
```diff
@@ -301,4 +306,65 @@ export class MessageModel extends DocumentDbModel<
...
+
+    // deserialize text into JSON
+    const contentOrError = t.validate(
+      JSON.parse(contentAsText),
+      MessageContent
+    );
+
+    if (isLeft(contentOrError)) {
+      return left<Error, Option<MessageContent>>(
+        new Error("Cannot deserialize stored message content")
```
cloudify: would be useful to append the validation error

lib/utils/objects.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Return an object filtering out keys that point to undefined values.
+ */
+export const definedValues = <T extends object>(obj: T): T =>
```
cloudify: name doesn't express the intent, perhaps `withOnlyDefinedValues` or `withoutUndefinedValues` would be more intuitive

lib/models/message.ts
gunzip: no, when the content is empty result = some("").

this could never happen in theory as the azure sdk api call should always returns a non null/undefined result.

lib/models/message.ts
gunzip: done

lib/models/message.ts
```diff
@@ -1,4 +1,5 @@
 import * as t from "io-ts";
+import { PathReporter } from "io-ts/lib/PathReporter";
```
cloudify: you can use [this reporter](https://github.com/teamdigitale/digital-citizenship-functions/blob/master/lib/utils/validation_reporters.ts#L42) that has a more readable output, it's the one we use for the API errors

PR #125, opened 2017-12-20
fix(#153802058): documentdb types

gunzip:
this avoids patching @types/documentdb

### DIFF
--- a/lib/types/documentdb.d.ts
+++ b/lib/types/documentdb.d.ts
@@ -1,0 +1,10 @@
+import * as documentdb from "documentdb";
+
+declare module "documentdb" {
+  // [#153802058] _ts is a number but is defined as string.
+  // TODO: get rid of this file when this pull request gets merged:
+  // https://github.com/DefinitelyTyped/DefinitelyTyped/pull/22291#issuecomment-352883251
+  export interface RetrievedDocument extends NewDocument, AbstractMeta {
+    _ts: any;
+  }
+}
--- a/lib/utils/documentdb.ts
+++ b/lib/utils/documentdb.ts
@@ -65,7 +65,7 @@ export const RetrievedDocument = t.intersection([
     _self: t.string,
 
     /** The time the object was created. */
-    _ts: t.string
+    _ts: t.union([t.string, t.number])
   }),
   t.partial({
     _attachments: t.string,


gunzip:
include in https://github.com/teamdigitale/digital-citizenship-functions/pull/124

lib/types/documentdb.d.ts
```diff
@@ -0,0 +1,10 @@
+import * as documentdb from "documentdb";
+
+declare module "documentdb" {
+  // [#153802058] _ts is a number but is defined as string.
+  // TODO: get rid of this file when this pull request gets merged:
```
cloudify: pls also open a story on pivotal

gunzip: we can leave the one referred in the comment above open:
https://www.pivotaltracker.com/story/show/153802058

cloudify: I would prefer to close that one (bug is fixed indeed) and add a different one as reminder... we don't know if/when they will merge it and I'd like to avoid keeping the bug story open...

lib/types/documentdb.d.ts
```diff
@@ -0,0 +1,10 @@
...
+  // https://github.com/DefinitelyTyped/DefinitelyTyped/pull/22291#issuecomment-352883251
+  export interface RetrievedDocument extends NewDocument, AbstractMeta {
```
cloudify: is this extending or replacing the definition of `RetrievedDocument` ?

gunzip: this overrides the type of _ts in RetrievedDocument (inherited from AbstractMeta)

lib/types/documentdb.d.ts
gunzip: ok

lib/types/documentdb.d.ts
cloudify: I see 👍 

and I guess the `declare module "documentdb" {}` statement, actually extends the module declaration from DefinitelyTyped? (not intuitive at all, it looks like an override)

gunzip: that's called https://www.typescriptlang.org/docs/handbook/declaration-merging.html

cloudify: brrr

PR #126, opened 2017-12-21
merged master and monkey patched url typings

gunzip:
if we want a cleaner workflow it's better to rebase your tech-debt branch on master


cloudify:
yes definitely, I'll rebase my branch after you merge the nodemailer typings...

PR #129, opened 2017-12-22
[#153811065] - mailup nodemailer integration

gunzip:


lib/utils/mailup.ts
```diff
@@ -0,0 +1,77 @@
...
+ * Implements MailUp REST API
+ */
+
+import * as request from "request";
+import * as winston from "winston";
+
+import { Either, isLeft, left, right } from "fp-ts/lib/Either";
+
+// request timeout in milliseconds
+const TIMEOUT = 10000;
```
cloudify: better name it `TIMEOUT_MS` then, always useful to keep info about metric together with var (wither with tagged type or else in var name)

lib/utils/mailup.ts
```diff
@@ -0,0 +1,77 @@
...
+
+type HttpMethod = "GET" | "POST";
+
+export interface ISmtpAuthInfo {
+  readonly Username: string;
```
cloudify: `NonEmptyString`?

lib/utils/mailup.ts
```diff
@@ -0,0 +1,77 @@
...
+import * as winston from "winston";
+
+import { Either, isLeft, left, right } from "fp-ts/lib/Either";
+
+// request timeout in milliseconds
+const TIMEOUT = 10000;
+
+type HttpMethod = "GET" | "POST";
+
+export interface ISmtpAuthInfo {
```
cloudify: if there's docs that explain these data structures, pls add a link to them

lib/utils/mailup.ts
```diff
@@ -0,0 +1,77 @@
...
+
+function callMailUpApi<T extends IApiResponse>(
+  method: HttpMethod,
+  url: string,
+  creds: ISmtpAuthInfo,
+  payload: {}
+): Promise<Either<Error, T>> {
+  return new Promise(resolve => {
+    const options = {
+      uri: url,
```
cloudify: probably makes sense to rename `url` to `uri`

lib/utils/mailup.ts
cloudify: also I guess this is actually a `DEFAULT_REQUEST_TIMEOUT_MS`

lib/utils/mailup.ts
```diff
@@ -0,0 +1,77 @@
...
+      method,
+      json: { ...payload, User: creds },
+      timeout: TIMEOUT
+    };
+    request(options, async (err, res, body) => {
```
cloudify: there's a Promise-enabled version of [request](https://www.npmjs.com/package/request#promises--asyncawait) (probably `request-promise-native` is what we need in this context)

lib/utils/mailup.ts
```diff
@@ -0,0 +1,77 @@
...
+      }
+      winston.error("callMailUpApiInternal|error|" + JSON.stringify(res));
+      return resolve(left<Error, T>(new Error("Cannot get access token")));
+    });
+  });
+}
+
+export async function sendTransactionalMail<T extends IApiResponse>(
+  creds: ISmtpAuthInfo,
+  payload: {}
```
cloudify: payload is an arbitrary object?

lib/utils/mailup_nodemailer.ts
```diff
@@ -0,0 +1,144 @@
...
+  options: IMailUpTransportOptions
+): nodemailer.Transport {
+  return {
+    name: TRANSPORT_NAME,
+
+    version: TRANSPORT_VERSION,
+
+    // tslint is wrong here, this *has* a typedef
+    // tslint:disable-next-line:typedef
+    async send(mail, callback) {
```
cloudify: perhaps tslint is happier with the form `send: async (mail, cb) => {}`

tsconfig.json
```diff
@@ -9,7 +9,6 @@
     "noUnusedParameters": true,
     "noImplicitReturns": true,
     "noFallthroughCasesInSwitch": true,
-    "strictFunctionTypes": true,
```
cloudify: why?

lib/utils/mailup.ts
```diff
@@ -0,0 +1,77 @@
...
+  readonly Secret: string;
+}
+
+export interface IApiResponse {
+  readonly Status: string;
+  readonly Code: string;
+  readonly Message: string;
+}
+
+function callMailUpApi<T extends IApiResponse>(
```
cloudify: not sure, why a type parameter is needed here?

lib/utils/mailup.ts
gunzip: I was thinking to replace request with superagent as it could be useful to write integration tests as well https://github.com/visionmedia/supertest

lib/utils/mailup.ts
gunzip: no, adding a type for it is on my todo list

tsconfig.json
gunzip: just to run test locally, I'll get rid of this once it's stable

lib/utils/mailup_nodemailer.ts
gunzip: yes =)

cloudify:
I think it makes sense to have this code in a dedicated repo, as it may be useful to others, what do you think?

gunzip:
that's right but atm the features are limited (to our scope). ie. there's no support for attachments, cc, bcc. probably it's not worth publish this as a standalone project without these features. moreover it's somewhat tight to our ecosystem (fp-ts, ReadableReporter...)

gulpfile.js
```diff
@@ -104,7 +106,7 @@ gulp.task("yarn:lint", () => {
 gulp.task("unit:test", () => {
   return gulp
     .src(TYPESCRIPT_SOURCE_DIR)
-    .pipe(run("jest --coverage --runInBand"));
+    .pipe(run("jest --coverage --runInBand --forceExit"));
```
cloudify: whis is `forceExit` needed?

lib/utils/mailup.ts
```diff
@@ -0,0 +1,287 @@
...
+import { ReadableReporter } from "./validation_reporters";
+
+// request timeout in milliseconds
+const DEFAULT_REQUEST_TIMEOUT_MS = 10000;
+
+// number of retries in case of HTTP errors
+const DEFAULT_REQUEST_RETRIES = 2;
+
+export const ENDPOINTS = {
+  sendTransactionalMail: "https://send.mailup.com/API/v2.0/messages/sendmessage"
```
cloudify: reason why this is not a constant by itself?

lib/utils/mailup.ts
```diff
@@ -0,0 +1,287 @@
...
+};
+
+type HttpMethod = "GET" | "POST";
+
+const TRANSPORT_NAME = "MailUp";
+const TRANSPORT_VERSION = "0.1";
+
+/**
+ * You need to create a SMTP+ user in MailUp administration panel
```
cloudify: link to the relevant doc?

lib/utils/mailup.ts
```diff
@@ -0,0 +1,287 @@
...
+}
+
+function callMailUpApi(
+  method: HttpMethod,
+  url: string,
+  creds: SmtpAuthInfo,
+  payload: {}
+): Promise<Either<Error, ApiResponse>> {
+  return request(method, url)
+    .timeout(DEFAULT_REQUEST_TIMEOUT_MS)
```
cloudify: what happens to the `Promise` when the req times our or max retries are reached?

lib/utils/mailup.ts
```diff
@@ -0,0 +1,287 @@
...
+  creds: SmtpAuthInfo,
+  payload: EmailPayload
+): Promise<Either<Error, ApiResponse>> {
+  const errorOrResponse = await callMailUpApi(
+    "POST",
+    ENDPOINTS.sendTransactionalMail,
+    creds,
+    payload
+  );
+  if (isLeft(errorOrResponse)) {
```
cloudify: you can use [Either#chain](https://github.com/gcanti/fp-ts/blob/master/src/Either.ts#L80)

lib/utils/mailup.ts
```diff
@@ -0,0 +1,287 @@
...
+      return left(
+        new Error(
+          `Error while sending email: ${response.Code} - ${response.Message}`
+        )
+      );
+    }
+  }
+}
+
+function toMailupAddresses(
```
cloudify: not super clear what this function and the next do, I would refactor them a little bit to make it easier to understand the validation logic - also functions that accept `undefined` values smell of problems (it's usually the caller that has to check the value and decide whether to call this function or not)

lib/utils/mailup.ts
```diff
@@ -0,0 +1,287 @@
...
+export function MailUpTransport(
+  options: IMailUpTransportOptions
+): nodemailer.Transport {
+  return {
+    name: TRANSPORT_NAME,
+
+    version: TRANSPORT_VERSION,
+
+    send: async (mail, callback) => {
+      // We don't use mail.data.from / mail.data.to as they are not parsed.
```
cloudify: I don't understand this statement: `as they are not parsed.` meaning they are ignored?

lib/utils/mailup.ts
```diff
@@ -0,0 +1,287 @@
...
+      // The following cast exists because of a bug in nodemailer typings.
```
cloudify: reference the bug (as it may be fixed in the future)

lib/utils/mailup.ts
```diff
@@ -0,0 +1,287 @@
...
+      const addresses: IAddresses = mail.message.getAddresses() as IAddresses;
+
+      // To convert nodemailer streams to strings:
```
cloudify: if this code is not needed, remove it

lib/utils/mailup.ts
```diff
@@ -0,0 +1,287 @@
...
+      //  const resolveContent = promisify.typed(mail.resolveContent);
+      //  const text = await resolveContent(mail.data, "text");
+      //  const html = await resolveContent(mail.data, "html");
+
+      const headers = Object.keys(mail.data.headers as {
```
cloudify: would be nice to describe every step in this code as it is not super clear how the send request is constructed and why is that way

gulpfile.js
gunzip: because for an unknow reason jest tests won't exit (at least on my win32 machine) when run through gulp (running the cli directly with node_modules/jest/bin/jest.js just works).
it could be a problem with gulp + run plugin + jest (pretty hard to debug, so I've just added the flag and it works)

lib/utils/mailup.ts
gunzip: because there are several endpoints in mailup api (ie. the endpoint to configure a new permitted sender address)

gulpfile.js
cloudify: then:

``` js
"jest --coverage --runInBand" + (process.platform === "win32" ? "--forceExit" : "")
```

+ add your comment above as explanation

lib/utils/mailup.ts
cloudify: so?

``` js
const SEND_TRANSACTIONAL_MAIL_ENDPOINT = "...";
const XYZ_ENDPOINT = "..."; 
```

or even better:

``` js
const SEND_TRANSACTIONAL_MAIL_ENDPOINT = process.env.MAILUP_SEND_TRANSACTIONAL_MAIL_ENDPOINT || "...";
// ...
```

lib/utils/mailup.ts
gunzip: no, it means that  mail.data.from and  mail.data.to are just "strings" which can represent invalid email addresses, so we use mail.message.getAddresses() instead to extract the already parsed addresses that nodemailer has validated for us.

lib/utils/mailup.ts
gunzip: is not clear from the nodemailer documentation if this is not needed. by now I've observed that 
mail.data.text and mail.data.subject are just plain text, but if you look at the nodemailer docs and typings it seems (but I couldn't verify it) that they CAN (in some **unspecified scenarios**) be streams that must be converted to strings before passing them to the MailUP API.

lib/utils/mailup.ts
gunzip: promise is reject with error (caught in catch())

cloudify: I think that from the caller point of view, this method should be very very simple, it should either send the email successfully or fail - once we have PR #133 merged, I don't see the point for having a different retry logic hidden here

cloudify: many reasons for not having a retry logic here:

* retry logic is opaque to the caller (what errors trigger a retry? parameters can't be configured)
* there's not logging whatsoever, the call can spend up to 10s to complete and we have no clue about what's going on inside (besides we decided to rely on message rescheduling to implement the retry mechanism to avoid long running message handlers)
* [the doc](http://visionmedia.github.io/superagent/#retrying-requests) says to use the retry logic only when the request is idempotent (definitely not our case)

gunzip: Agreed and removed the retry line. Note that the retry here was harmless has it would have occurred only on timeouts and server errors, moreover most of the statements above (ie. retry opaque to caller) apply to the azure api calls as well, as they use a [retry policy](https://github.com/Azure/autorest/blob/master/docs/client/retry.md) similar to the one implemented by superagent.

lib/utils/mailup.ts
gunzip: added

PR #131, opened 2017-12-27
[#153722616]  feat: strict model validation

gunzip:
reject unknow object properties in t.validate()
see https://github.com/gcanti/io-ts/issues/106


lib/utils/objects.ts
```diff
@@ -1,19 +1,53 @@
+import * as t from "io-ts";
+
+export const isObject = (o: {}) =>
```
cloudify: btw `utils/types` is meant to be a container for all type-related function (e.g. type checking, io-ts helpers, etc...) pls consider moving everything in there to avoid confusion

lib/utils/objects.ts
```diff
@@ -1,19 +1,53 @@
...
-              typeof value === "object" ? withoutUndefinedValues(value) : value
+            [k]: isObject(value) ? withoutUndefinedValues(value) : value
           }
         : acc;
     },
     {} as T
   ) as T;
+}
+
+export function strictInterfaceWithOptionals<
```
cloudify: +function comment

lib/utils/validation_reporters.ts
```diff
@@ -14,6 +14,10 @@ function getContextPath(context: Context): string {
   const keysPath = context.map(({ key }) => key).join(".");
   const lastType = context[context.length - 1].type;
 
+  if ("never" === lastType.name) {
```
cloudify: perhaps add a [test case](https://github.com/teamdigitale/digital-citizenship-functions/blob/master/lib/utils/__tests__/validation_reporters.test.ts) for this

cloudify:
general comment, with the current subject, Pivotal will not link the commit to the story, for the automatic link to work, the story ID must be in square brackets ad the beginning of the commit message, i.e.: `[#153722616] ....`

lib/utils/validation_reporters.ts
gunzip: done

lib/utils/types.ts
```diff
@@ -90,7 +90,7 @@ export function withoutUndefinedValues<T extends object>(obj: T): T {
 }
 
 /**
- *  Returns a new type that validates succesfully only
+ *  Return a new type that validates successfully only
```
cloudify: `Returns` is actually better that `Return` (the comment refers to the function)

PR #132, opened 2017-12-30
Fix IResponse types

cloudify:
While I was writing an article describing the technique we implemented, I realized the code I wrote wasn't correct, `IResponse` was defined with the `kind` attribute of type `string` when instead the `kind` attribute has the type that names the kind of response (e.g. `IResponseNotFound`), `IResponse` should have a type parameter that is the type of `kind`.

### DIFF
--- a/lib/utils/response.ts
+++ b/lib/utils/response.ts
@@ -19,8 +19,8 @@ const HTTP_STATUS_500 = 500 as HttpStatusCode;
  * Interface for a Response that can be returned by a middleware or
  * by the handlers.
  */
-export interface IResponse {
-  readonly kind: string;
+export interface IResponse<T> {
+  readonly kind: T;
   readonly apply: (response: express.Response) => void;
 }
 
@@ -31,8 +31,8 @@ export interface IResponse {
 /**
  * Interface for a successful response returning a json object.
  */
-export interface IResponseSuccessJson<T> extends IResponse {
-  readonly kind: "IResponseSuccessJson";
+export interface IResponseSuccessJson<T>
+  extends IResponse<"IResponseSuccessJson"> {
   readonly value: T; // needed to discriminate from other T subtypes
 }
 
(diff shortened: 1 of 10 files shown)


lib/utils/response.ts
```diff
@@ -19,8 +19,8 @@ const HTTP_STATUS_500 = 500 as HttpStatusCode;
  * Interface for a Response that can be returned by a middleware or
  * by the handlers.
  */
-export interface IResponse {
-  readonly kind: string;
+export interface IResponse<T> {
+  readonly kind: T;
```
gunzip: would it be useful to use an enum here (T extends MyEnum) with all the tags ? 
maybe to avoid typos when using the IResponses

gunzip: T can be anything (or extends string) ?

cloudify: looks like literal types can be a strings, numbers or booleans (see [discriminated unions](https://basarat.gitbooks.io/typescript/docs/types/discriminated-unions.html)) - if you misspell the type you should get a build error:

``` ts
interface IResponse<T> { 
    kind: T;
};

const a: IResponse<"A"> = {
    kind: "B", // error
};
```

gunzip: so shouldn't be 

```
export interface IResponse<T extends string>
```

moreover, I'm concerned about typos like

```
export function ContextMiddleware<T>(): IRequestMiddleware<"FooBar">
```

is there any check that 'FooBar' does exists indeed ?

cloudify: The point is not that `"FooBar"` exists, the point is that `IRequestMiddleware<"FooBar", T>` can only produce `IResponse<"FooBar">` responses.

cloudify: Try to play with the code and see by yourself that this refactoring doesn't lose any type safety.

gunzip: it looks like it works with const enum literals as discriminator. that would be good to keep intellisense autocomplete on kind(s) property.

cloudify: I'll merge this in the meantime, then you can do the PR with the enums

PR #133, opened 2017-12-31
[#150597257] - schedule retries when a queue action fails

gunzip:
Work in progress for https://www.pivotaltracker.com/story/show/150597257

We can re-schedule a message in the queue for later processing using visibilityTimeout.

Two possible retry stategies: 

1. **update** the visibilityTimeout of the received message then put it again in the queue. this strategy use the `dequeCount` property in Functions bindings and calls `context.done(err)` to return the message to the queue. after 7 days the message will be deleted from the queue in any case.

2. **create** a new message in the queue with the computed visibilityTimeout. this strategy does not rely on `dequeCount` (as it's always = 1) and calls `context.done()` to remove the old message from the queue. the number of retries is saved in the message payload.

## 1 Pros (update the same message)

- when maxDequeCount will be reached the message is moved automatically into the poison-queue
- you don't have to save the number of retries into the message payload (as dequeCount is a system property)

## 2 Pros (create a new message for every retry and delete the old one)

- you can check the number of retries looking into the message payload
- number of retries does not depened on maxDequeCount 
- you can schedule retries for more than 7 days
- context.done() does not log an error

Probably #1 is more useful if you plan to use the poison queue someway.

I chose to not use an exponential backoff strategy because the intervals between retry (in seconds) where not optimal. using a fixed number of intervals gives us more control.

see https://github.com/teamdigitale/digital-citizenship-functions/pull/133/files#diff-2d141546a695a0fa4638dedaf20abc72R9

### DIFF
--- a/lib/utils/azure_queues.ts
+++ b/lib/utils/azure_queues.ts
@@ -1,0 +1,111 @@
+/**
+ * Implements utility functions for Azure Storage queues
+ */
+import * as winston from "winston";
+
+import { IContext } from "azure-functions-types";
+import { QueueService } from "azure-storage";
+
+import { Option, some } from "fp-ts/lib/Option";
+
+// Any delay must be less than 7 days (< 604800 seconds)
+const MAX_BACKOFF_MS = 7 * 24 * 3600 * 1000;
+const MIN_BACKOFF_MS = 285;
+
+// MAX_RETRIES *must* equal (maxDequeueCount - 1)
+// if (MAX_RETRIES < maxDequeueCount) - 1 there will be other extraneous
+//    #(maxDequeueCount - MAX_RETRIES) retries with the default visibilityTimeout
+//    before putting the message into the poison queue
+// if (MAX_RETRIES > maxDequeueCount - 1) the system will retry even after maxDequeueCount
+//    is reached and duplicate messages will be put into the poison queue
+export const MAX_RETRIES = Math.floor(
+  Math.log2(MAX_BACKOFF_MS / MIN_BACKOFF_MS)
+);
+
+/**
+ * Compute the timeout in seconds before the message will be processed again.
+ * returns none in case the maximum number of retries is reached
+ */
+export const getDelaySecForRetries = (
+  retries: number,
+  maxRetries = MAX_RETRIES,
+  minBackoff = MIN_BACKOFF_MS
+): Option<number> =>
+  some(retries)
+    .filter(nr => nr <= maxRetries)
+    .map(nr => Math.ceil(minBackoff * Math.pow(2, nr) / 1000));
+
+export function queueMessageToString(context: IContext): string {
+  /* istanbul ignore next */
+  return [
+    "bindings = ",
+    JSON.stringify(context.bindings),
+    "; queueTrigger = ",
+    context.bindingData.queueTrigger,
+    "; expirationTime = ",
+    context.bindingData.expirationTime,
+    "; insertionTime = ",
+    context.bindingData.insertionTime,
+    "; nextVisibleTime = ",
+    context.bindingData.nextVisibleTime,
+    "; id = ",
+    context.bindingData.id,
+    "; popReceipt = ",
+    context.bindingData.popReceipt,
+    "; dequeueCount = ",
+    context.bindingData.dequeueCount
+  ].join("");
+}
+
+/**
+ * Schedule the re-processing of a message in the queue
+ * after an incremental delay (visibilityTimeout).
+ *
+ * Useful in case of transient errors. The message is enqueued with
+ * a newly computed visibilityTimeout (proportional to dequeueCount)
+ *
+ * @param queueService  The Azure storage queue service
+ * @param queueName     The Azure storage queue name
+ * @param context       The Functions context with bindings
+ */
+export function retryMessageEnqueue(
+  queueService: QueueService,
+  queueName: string,
+  context: IContext
+): void {
+  const queueMessage = context.bindingData;
+
+  winston.debug(`Retry to handle message ${queueMessageToString(context)}`);
+
+  // dequeueCount starts with one (not zero)
+  const numberOfRetries = queueMessage.dequeueCount;
+
+  getDelaySecForRetries(numberOfRetries)
+    .map(visibilityTimeoutSec => {
+      // update message visibilityTimeout
+      queueService.updateMessage(
+        queueName,
+        queueMessage.id,
+        queueMessage.popReceipt,
+        visibilityTimeoutSec,
+        err => {
+          context.done(
+            err ||
+              `Retrying with timeout|retry=${numberOfRetries}|timeout=${visibilityTimeoutSec}|queueMessageId=${
+                queueMessage.id
+              }`
+          );
+        }
+      );
+    })
+    .getOrElse(() => {
+      winston.info(
+        `Maximum number of retries reached|retries=${numberOfRetries}|${
+          queueMessage.id
+        }`
+      );
+      context.done(
+        `Moving queueMessage=${queueMessage.id} to ${queueName}-poison`
+      );
+    });
+}
--- a/lib/utils/errors.ts
+++ b/lib/utils/errors.ts
@@ -1,0 +1,37 @@
+/**
+ * Contains Error types,
+ * used to clarify the intent of a throw
+ * and useful inside unit tests.
+ */
+
+export const enum ErrorTypes {
+  TransientError = "TransientError",
+  PermanentError = "PermanentError"
+}
+
+interface IRuntimeError<T extends ErrorTypes> {
+  readonly kind: T;
+  readonly message: string;
+  readonly cause?: Error;
+}
+
+const RuntimeError = <T extends ErrorTypes>(
+  kind: T
+): ((message: string, cause?: Error) => IRuntimeError<T>) => {
+  return (message: string, cause?: Error) => ({
+    cause,
+    kind,
+    message
+  });
+};
+
+export type TransientError = IRuntimeError<ErrorTypes.TransientError>;
+export const TransientError = RuntimeError(ErrorTypes.TransientError);
+
+export type PermanentError = IRuntimeError<ErrorTypes.PermanentError>;
+export const PermanentError = RuntimeError(ErrorTypes.PermanentError);
+
+export type RuntimeError = TransientError | PermanentError;
+
+export const isTransient = (error: RuntimeError): error is TransientError =>
+  error.kind === ErrorTypes.TransientError;
--- a/host.json
+++ b/host.json
@@ -6,6 +6,13 @@
     "maxConcurrentRequests": 50,
     "dynamicThrottlesEnabled": false
   },
+  "queues": {
+    "maxPollingInterval": 10000,
+    "visibilityTimeout" : "00:00:30",
+    "batchSize": 16,
+    "maxDequeueCount": 22,
+    "newBatchThreshold": 8
+  },
   "functions": [
     "PublicApiV1",
     "Openapi",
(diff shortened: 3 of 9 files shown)


lib/utils/azure_queues.ts
```diff
@@ -0,0 +1,100 @@
...
+
+// Array of intervals, in seconds, to delay message processing
+// The array *must* contain = (maxDequeueCount - 1) items
+// Any value must be less than 7 days (< 604800 seconds)
+// if (MAX_RETRIES < maxDequeueCount) - 1 there will be other extraneous
+//    #(maxDequeueCount - MAX_RETRIES) retries with the default visibilityTimeout
+//    before putting the message into the poison queue
+// if (MAX_RETRIES > maxDequeueCount - 1) the system will retry even after maxDequeueCount
+//    is reached and duplicate messages will be put into the poison queue
+export const RETRIES: ReadonlyArray<number> = [
```
cloudify: where do these numbers come from? truncated exp backoff is perfectly fine, see an example implementation [here](https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala#L149)

btw you should add a randomized factor to the delay - see page 285 of [this classic school text](http://www.uoitc.edu.iq/images/documents/informatics-institute/exam_materials/Computer%20Networks%20-%20A%20Tanenbaum%20-%205th%20edition.pdf) for an explanation

lib/utils/errors.ts
```diff
@@ -0,0 +1,41 @@
...
+  PermanentError = "PermanentError"
+}
+
+interface ICustomError<T extends ErrorTypes> {
+  readonly kind: T;
+  readonly message: string;
+  readonly cause?: Error;
+}
+
+const CustomError = <T extends ErrorTypes>(
```
cloudify: not sure why you need the type parameter `T`, I don't see any code that needs to be generic in `T` - it looks like the following would be enough:

``` ts
interface ICustomError {
   readonly kind: ErrorTypes;
   readonly message: string;
   readonly cause?: Error;
 }
```

gunzip: ok

lib/utils/azure_queues.ts
gunzip: Using Math.min(maxBackoff, Math.pow(10, retries)) was [my first try](https://github.com/gunzip/digital-citizenship-functions/blob/backoff/lib/created_message_queue_handler.ts#L104).

With this formula you get retries at [ 10, 100, 1000, 10000, 100000 ], so the last retry is at about 1.5 days from the moment the message has been enqueued. So I began tweaking the formula to get more retries, considering that we have a timespan of 7 days (max time_to_live) for each message. Finally I asked myself "how using a (tweaked) formula is better than directly specify the intervals directly ?" and went with the current implementation. 

The random delay is useful if you must avoid event collisions which is not our case IMHO: I'm not biased against adding a random factor, but note that tweaking the visibilityTimeout does not guarantee that the message will be scheduled at that exact time in the future, it only ensures that the message won't be processed *before the specified delay* (with some kind of imprecision dued to the Functions runtime). The moment the message will be dequeued depends on backpressure as well.

gunzip:
as the [issue regarding time_to_live](https://www.pivotaltracker.com/story/show/150588117) is strictly related to this one regarding scheduled retries I was planning to solve both with one PR. At some point in time we must avoid to schedule retries that exceed the ttl.

lib/utils/azure_queues.ts
cloudify: untitled

cloudify: `300ms * 2^count` looks quite reasonable

lib/utils/errors.ts
cloudify: maybe needed in the future === not needed 😸

lib/utils/azure_queues.ts
gunzip: 300ms * 2^count translates into 6 retries in the first 20 seconds (aka 6 more messages to process in the queue fo each message that end up with an error). IMHO this is way too much for the kind of errors that could cause a retry (network errors, paas service downtime, etc.) which are probably restored in seconds (if not minutes or hours). moreover this will pollute the queue with a lot of messages in case of subsequent errors. 

finally, another issue with any exponential strategy is that your last retry will be at best at 3.5 days from the time the message is accepted. this means, according to our max timespan of 7 days, you'll lose 3.5 days of possible retries (for example in case we opt for 300ms^2*count the last retry is at 3.6 days, then the message will expire)

cloudify: I think we are in the territory of "opinions" and unless you have some hard data to back your assumptions I would go for the [industry standard approach](https://docs.aws.amazon.com/general/latest/gr/api-retries.html) :)

cloudify: does `285ms * (2 ^ count)` look better?

untitled

gunzip: nicer ! BUT it looks like the visibilityTimeout granularity is in seconds (int):
http://azure.github.io/azure-storage-node/QueueService.html#updateMessage__anchor
so we must round the values or we have to chose a totally different way

cloudify: just round by excess using `Math.ceil()`

gunzip: I've plugged in the [backoff strategy](https://github.com/teamdigitale/digital-citizenship-functions/pull/133/files#diff-2d141546a695a0fa4638dedaf20abc72R23). 

Now we have ca. 30 retries. As maxDequeCount must be configured in host.json globally (once for every queue handler), it means that maxDequeCount must be 31, this in case you use a retry strategy OR NOT (so you'd better do : ). 

TL; DR: If you don't call retry() in your queue handlers they will retry enqueing the message 30 times in a very fast burst.

lib/utils/azure_queues.ts
```diff
@@ -6,6 +6,12 @@ import * as winston from "winston";
 import { IContext } from "azure-functions-types";
 import { QueueService } from "azure-storage";
 
+import { unfoldr } from "fp-ts/lib/Array";
+import { none, Option, some } from "fp-ts/lib/Option";
+
+const TIME_TO_LIVE = 7 * 24 * 3600 * 1000;
+const MILLISECONDS = 285;
```
cloudify: `MIN_BACKOFF_MS`

lib/utils/azure_queues.ts
```diff
@@ -6,6 +6,12 @@ import * as winston from "winston";
 import { IContext } from "azure-functions-types";
 import { QueueService } from "azure-storage";
 
+import { unfoldr } from "fp-ts/lib/Array";
+import { none, Option, some } from "fp-ts/lib/Option";
+
+const TIME_TO_LIVE = 7 * 24 * 3600 * 1000;
```
cloudify: `MAX_BACKOFF_MS`

lib/utils/azure_queues.ts
```diff
@@ -14,16 +20,14 @@ import { QueueService } from "azure-storage";
...
-  21600,
-  86400,
-  172800,
-  345600
-];
+export const RETRIES: ReadonlyArray<number> = unfoldr((retries: number): Option<
+  [number, number]
+> => {
+  const delay = Math.ceil(MILLISECONDS * Math.pow(2, retries) / 1000);
+  return delay <= TIME_TO_LIVE
```
cloudify: `TIME_TO_LIVE` is also in msec

lib/utils/azure_queues.ts
```diff
@@ -14,16 +20,14 @@ import { QueueService } from "azure-storage";
...
-  10,
-  60,
-  600,
-  3600,
-  21600,
-  86400,
-  172800,
-  345600
-];
+export const RETRIES: ReadonlyArray<number> = unfoldr((retries: number): Option<
```
cloudify: you should make this a function now `(retry: number) => number`

gunzip: I need to know in advance how many retries can be scheduled in the the specified time to live (RETRIES.length) to check that maxDequeCount is setup correctly in hosts.json

cloudify: `max_retries < log_2(TIME_TO_LIVE / MIN_BACKOFF)`

cloudify: `max_retries = [ log_2(7 * 24 * 3600 * 1000 / 285) ] = 21`

perhaps there's a 1 off difference from your implementation, anyway it's simple algebra, you don't have to precalculate all values

gunzip: of course. it was just to don't have to write the logic (pow) and the reverse logic (log) and then keep them in sync (I'll turn it to a function anyway)

lib/utils/azure_queues.ts
```diff
@@ -0,0 +1,109 @@
...
+//    #(maxDequeueCount - MAX_RETRIES) retries with the default visibilityTimeout
+//    before putting the message into the poison queue
+// if (MAX_RETRIES > maxDequeueCount - 1) the system will retry even after maxDequeueCount
+//    is reached and duplicate messages will be put into the poison queue
+export const MAX_RETRIES = Math.floor(
+  Math.log2(MAX_BACKOFF_MS / MIN_BACKOFF_MS)
+);
+
+/**
+ * Compute the timeout in seconds before the message will be processed again
```
cloudify: Add a note about the meaning of an empty option

lib/utils/azure_queues.ts
```diff
@@ -0,0 +1,109 @@
...
+ * after an incremental delay (visibilityTimeout).
+ *
+ * Useful in case of transient errors. The message is enqueued with
+ * a newly computed visibilityTimeout (proportional to dequeueCount)
+ *
+ * @param queueService  The Azure storage queue service
+ * @param queueName     The Azure storage queue name
+ * @param context       The Functions context with bindings
+ */
+export function retry(
```
cloudify: `retry` is a pretty overloaded identifier, consider perhaps a more specific identifier (e.g. `retryMessage` or a more explicit one `retryAfterDelay`)

cloudify: `retry` is a pretty overloaded identifier, consider perhaps a more specific identifier (e.g. `retryMessage` or a more explicit one `retryAfterDelay`)

lib/utils/azure_queues.ts
```diff
@@ -0,0 +1,109 @@
...
+// if (MAX_RETRIES > maxDequeueCount - 1) the system will retry even after maxDequeueCount
+//    is reached and duplicate messages will be put into the poison queue
+export const MAX_RETRIES = Math.floor(
+  Math.log2(MAX_BACKOFF_MS / MIN_BACKOFF_MS)
+);
+
+/**
+ * Compute the timeout in seconds before the message will be processed again
+ */
+export const delay = (
```
cloudify: `delay` is usually used to delay an actual amount of time, perhaps `getDelayMsForRetries` will be less confusing

lib/utils/azure_queues.ts
cloudify: Also in the comment you're saying "seconds" whereas in the body you're using `MIN_BACKOFF_MS` without dividing by 1000 (an actual `Duration` type would be useful here) :)

lib/utils/azure_queues.ts
```diff
@@ -0,0 +1,109 @@
...
+  context: IContext
+): void {
+  const queueMessage = context.bindingData;
+
+  winston.debug(`Retry to handle message ${queueMessageToString(context)}`);
+
+  const numberOfRetries = queueMessage.dequeueCount;
+
+  delay(numberOfRetries)
+    .map(visibilityTimeout => {
```
cloudify: when you're dealing with `number` variables that represent durations it's always useful to have the unit of measure as part of the variable name, e.g. `visibilityTimeoutSec` (or `visibilityTimeoutMs`?)

lib/utils/errors.ts
cloudify: I would rename `CustomError` to `RuntimeError`, seems more effective in communicating the intent

PR #136, opened 2018-01-04
[#154066374] fix: enumtype

gunzip:


lib/utils/types.ts
```diff
@@ -14,22 +14,15 @@ export type Tagged<T, S, A> = t.Type<S, A & T>;
...
 /**
  * Creates an io-ts Type from an enum
  */
 export const enumType = <E>(e: {}, name: string): t.Type<any, E> => {
-  const values = getObjectValues(e);
-
   return new t.Type<any, E>(
     name,
-    (v): v is E => typeof v === "string" && values.indexOf(v) >= 0,
-    (s, c) => (this.is(s) ? t.success(s) : t.failure(s, c)),
```
gunzip: "this.is is not a function"

lib/utils/types.ts
```diff
@@ -14,22 +14,15 @@ export type Tagged<T, S, A> = t.Type<S, A & T>;
...
-  Object.keys(obj).reduce<ReadonlyArray<string>>(
-    (acc, key) => [...acc, (obj as any)[key]],
-    []
-  );
-
 /**
  * Creates an io-ts Type from an enum
  */
 export const enumType = <E>(e: {}, name: string): t.Type<any, E> => {
-  const values = getObjectValues(e);
```
gunzip: why not just check for e[v] ?

cloudify: The [Object.keys()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys) method returns an array of a given object's own enumerable properties, skipping inherited properties. 

For instance `__proto__` (an inherited property) is a valid property for `e` but not a valid enum value.

lib/utils/types.ts
cloudify: ops yes, `this` is not accessible in this scope, I guess we could as well extract the `is` logic to a function:

```ts
const isE: (v: any) => boolean = v =>
    typeof v === "string" && values.indexOf(v) >= 0;

return new t.Type<any, E>(
  name,
  (v: any): v is E => isE(v),
  (s, c) => (isE(s) ? t.success(s) : t.failure(s, c)),
  t.identity
);
```

lib/utils/types.ts
gunzip: I've added a check with hasOwnProperties()

lib/utils/types.ts
gunzip: done

lib/utils/types.ts
gunzip: Sorry only now I realized that you're validating the value and not the key, I'll revert it back :)

PR #141, opened 2018-01-21
[#154335393] GetService in public API

gunzip:


api/definitions.yaml
```diff
@@ -241,6 +241,28 @@ Service:
...
+    service_id:
+      $ref: "#/ServiceId"
+    service_name:
+      $ref: "#/ServiceName"
+    organization_name:
+      $ref: "#/OrganizationName"
+    department_name:
+      $ref: "#/DepartmentName"
+    version:
+      type: integer
```
cloudify: is it possible to define a non-negative integer with swagger constrainta?

api/definitions.yaml
```diff
@@ -241,6 +241,28 @@ Service:
...
+    id:
```
cloudify: the actual document ID is redundant as it is composed of `service_id` and `version` and should be used internally only

api/public_api_v1.yaml
```diff
@@ -9,6 +9,33 @@ schemes:
...
+      summary: Get Service
+      description: A previously created service with the provided service ID is returned.
+      responses:
+        '200':
+          description: Service found.
+          schema:
+            "$ref": "#/definitions/ServicePublic"
+          examples:
+            application/json:
+              id: 2b3e728c1a5d1efa035c-0000000000000001
```
cloudify: see my comment above

lib/api/admin_api.ts
```diff
@@ -497,6 +497,25 @@ export const specs = {
         "authorized_recipients"
       ]
     },
+    ServicePublic: {
+      title: "Service (public)",
+      description: "A Service tied to an user's subscription.",
```
cloudify: `tied` has a literal meaning (having one's hands tied), perhaps `associated` would be better in this context

lib/api/admin_api.ts
```diff
@@ -497,6 +497,25 @@ export const specs = {
...
+      type: "object",
+      properties: {
+        service_id: { $ref: "#/definitions/ServiceId" },
+        service_name: { $ref: "#/definitions/ServiceName" },
+        organization_name: { $ref: "#/definitions/OrganizationName" },
+        department_name: { $ref: "#/definitions/DepartmentName" },
+        version: { type: "integer" },
+        id: { type: "string" }
```
cloudify: see my comment above

lib/controllers/services.ts
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * Converts a retrieved service to a service that can be shared via API
+ */
+function retrievedServiceToPublic(
+  retrievedService: RetrievedService
+): ApiService {
+  return {
+    department_name: retrievedService.departmentName,
+    id: retrievedService.id,
```
cloudify: see my comment above

lib/controllers/services.ts
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Wraps a GetService handler inside an Express request handler.
+ */
+export function GetService(serviceModel: ServiceModel): express.RequestHandler {
+  const handler = GetServiceHandler(serviceModel);
+  const azureUserAttributesMiddleware = AzureUserAttributesMiddleware(
+    serviceModel
+  );
+  const middlewaresWrap = withRequestMiddlewares(
+    AzureApiAuthMiddleware(new Set([UserGroup.ApiServiceRead])),
```
cloudify: I think we should differentiate the permissions for the admin endpoint and the public endpoint, we should have an `ApiPublicServiceRead` permission

api/definitions.yaml
gunzip: yes but then we have to put this constraint on all the non negative integer fields which is probably beyond this PR scope

cloudify: agreed

PR #144, opened 2018-02-01
Upgrades Typescript to 2.7.1

cloudify:


cloudify:
uhm seems like there's a problem with one test that never ends, possibly `emailnotifications_queue_handler.test.ts`

PR #148, opened 2018-02-11
[#155042061]: refactor notification status

gunzip:
## Context

At the moment:

1. we don't have a status field for messages (only for notifications), but messages can be throttled and/or can fail as well, so we have to to plan a status model (this states seamlessy for read / unread status as described in https://www.pivotaltracker.com/story/show/150737636)

2. we don't have a status field for notifications. notifications models are updated in-place without being versioned

3. we don't have a mechanism to handle multiple notifications (channels) statuses

4. the notification actually carries metadata information for *all* channels so we have to check for undefined fields (ie. emailNotification) in the document object to know which kind of notification we are processing (which is a little bit ugly). moreover this is very bad for concurrency as our
update() method for versioned models has a race-condition (between find() and create()), so it's better to avoid concurrent updates of the same models.

## Aim of this PR

- make notification model immutable and avoid in-place update (to set the status)
- add a new 'notification status' model (one-to-one with notification, but versioned)
- add different notification types (by now: EmailNotification)
- avoid concurrent updates from different channel handlers of the same documents
- make it easier to loop on notification channels and statuses

See

- [notification.ts](https://github.com/teamdigitale/digital-citizenship-functions/pull/148/files#diff-f77a48800133699f87c40a0d01fef39e)
- [messages.ts](https://github.com/teamdigitale/digital-citizenship-functions/pull/148/files#diff-ca9342512f1db78ecb71b1a86cd82b77)

The implemented model is so far:

`message (immutable) ---- 1-to-1 ----> notification (immutable, one document for all channel) ---- 1-to-n ----> notification_status (versioned, one for each channel)`

The NotificationStatus id is computed as notificationId:channel:

```
{
    "channel": "EMAIL",
    "messageId": "01C79QW09FPMR0468NXET2CF6X",
    "notificationId": "01C79Q0Z9YT0W06WCRZ6PJ9PWS",
    "status": "SENT_TO_CHANNEL",
    "statusId": "01C79Q0Z9YT0W06WCRZ6PJ9PWS:EMAIL",
    "updateAt": "2018-02-26T19:02:18.767Z",
    "id": "01C79QWZ9YT0WA6WCRZ6PJ9PWS:EMAIL-0000000000000000",
    "version": 0,
    ...
}
```

for message we'll have (TBD)

`message (immutable) ---- 1-to-1 ----> message_validity_status (versioned)`
`message (immutable) ---- 1-to-1 ----> message_read_status (versioned)`


cloudify:
please split this huge PR in multiple PRs (I suggest the different goals in "Aim of this PR" should be separate PRs)

gunzip:
that would be pretty difficult as all the aims imply adding a new model (notification_status).
this changes the API specs (to get the models right), message controller (to get notification status) and queue handlers (to set notification status) and all the relative tests.

the only things I see could be splitted out are the mods to DAO (documentdb_model_*), a new task (in tasks.json) and the upgrade to io-ts, but they are minimal changes.

lib/controllers/messages.ts
```diff
@@ -197,6 +201,75 @@ type IGetMessagesHandler = (
...
+ * to retrieve the relative notification status.
+ *
+ * @returns an object with channels as keys and statuses as values
+ *          ie. { email: "SENT_TO_CHANNEL" }
+ */
+async function getMessageNotificationStatuses(
+  notificationModel: NotificationModel,
+  notificationStatusModel: NotificationStatusModel,
+  retrievedMessage: RetrievedMessage
+): Promise<Either<Error, Promise<NotificationStatusHolder>>> {
```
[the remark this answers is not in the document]
cloudify: I have refactored the logic of this method [here](https://github.com/gunzip/digital-citizenship-functions/pull/1/files), I believe it's more easier to follow - I've used `Promise.all` to collect the results of all channel statuses.

lib/controllers/messages.ts
```diff
@@ -197,6 +201,75 @@ type IGetMessagesHandler = (
...
+    // It may happen that the notification object is not yet created in the database
+    // due to some latency, so it's better to not fail here but return an empty object
+    const maybeNotification = errorOrMaybeNotification.value;
+    if (isNone(maybeNotification)) {
+      winston.info(
+        `getMessageNotificationStatuses|Notification not found|messageId=${
+          retrievedMessage.id
+        }`
+      );
+      return right(Promise.resolve({}));
```
cloudify: I think it makes sense to have an `Option` here to reflect the semantic of `maybeNotification` (that is an `Option` as well), consider having this method return a `Promise<Either<Error, Option<NotificationStatusHolder>>>`

lib/controllers/messages.ts
```diff
@@ -197,6 +201,75 @@ type IGetMessagesHandler = (
...
+          statusesP: Promise<NotificationStatusHolder>,
+          channelIdx: string
+        ): Promise<NotificationStatusHolder> => {
+          const channel =
+            NotificationChannelEnum[channelIdx as NotificationChannelEnum];
+          const errorOrMaybeStatus = await notificationStatusModel.findOneNotificationStatusByNotificationChannel(
+            notification.id,
+            channel
+          );
+          const statusObj = fromEither(errorOrMaybeStatus)
```
cloudify: I'm not super happy about swallowing errors, can we have `NotificationStatusHolder` as the following and let the caller decide to consider or ignore the errors?

``` ts
NotificationStatusHolder {
   readonly [channel in NotificationChannelEnum]: Either<Error, NotificationChannelStatusValueEnum>
}
```

lib/created_message_queue_handler.ts
```diff
@@ -112,34 +111,73 @@ type ContextWithBindings = t.TypeOf<typeof ContextWithBindings> & IContext;
  * @param maybeProfile      the recipient's profile (or none)
  * @param defaultAddresses  default addresses (one per channel)
  */
-function getEmailNotification(
+function tryGetDestinationEmail(
   maybeProfile: Option<RetrievedProfile>,
```
cloudify: pls add a TODO here, we need to refactor this method and the next one, methods shouldn't expect `Option` as parameters, this is almost always a code smell that indicated a leakage of control flow from the caller to the callee, it should be responsibility of the caller to check the content of an `Option`

lib/created_message_queue_handler.ts
```diff
@@ -112,34 +111,73 @@ type ContextWithBindings = t.TypeOf<typeof ContextWithBindings> & IContext;
...
+/**
+ * Save a notification object to the database.
+ *
+ * @param notificationModel notification model used to interact with database
+ * @param notifications     array of notification objects
+ */
+async function saveNotification(
+  notificationModel: NotificationModel,
+  notification: NewNotification
+): Promise<Either<RuntimeError, RetrievedNotification>> {
```
cloudify: since this always errors with `TransientError` you can be more specific here and return `Promise<Either< TransientError, RetrievedNotification>>` (if possible, it's always good to hint the compiler about more specific types)

lib/created_message_queue_handler.ts
```diff
@@ -246,39 +269,46 @@ export function processResolve(
...
-    // we output a notification event to the email channel queue
-    if (notification.emailNotification) {
-      // tslint:disable-next-line:no-object-mutation
-      context.bindings.emailNotification = {
-        messageContent,
-        messageId: notification.messageId,
-        notificationId: notification.id,
-        senderMetadata
-      };
+  if (isLeft(errorOrNotification)) {
```
cloudify: minor style improvement, just after you verify that the `Either` is a left, you can extract its left value, i.e.:

```ts
if (isLeft(errorOrNotification)) { 
  const error = errorOrNotification.value;
  ...
```

and you'll save the duplicated assignment (274 and 286)

lib/models/notification.ts
```diff
@@ -85,6 +103,23 @@ export const NewNotification = tag<INewNotificationTag>()(
 
 export type NewNotification = t.TypeOf<typeof NewNotification>;
 
+/**
+ * Factory method to make NewNotification objects
+ */
+export function makeNotification(
```
cloudify: `createNewNotification` ?

lib/models/notification.ts
```diff
@@ -100,11 +135,7 @@ export const RetrievedNotification = tag<IRetrievedNotificationTag>()(
 export type RetrievedNotification = t.TypeOf<typeof RetrievedNotification>;
 
 function toBaseType(o: RetrievedNotification): Notification {
-  return {
-    emailNotification: o.emailNotification,
-    fiscalCode: o.fiscalCode,
-    messageId: o.messageId
-  };
+  return pick(["fiscalCode", "messageId", "channels"], o);
```
cloudify: nop, you lose type safety here

lib/utils/types.ts
```diff
@@ -2,6 +2,22 @@ import * as t from "io-ts";
 
 import { Set as SerializableSet } from "json-set-map";
 
+/**
+ * Returns a subset of the input objects fields.
+ *
+ * @param obj the input object
+ * @param props an array of keys to preserve
+ */
+export const pick = <T, K extends keyof T>(
```
cloudify: I don't like this method as it loses type information, in general I'm very much against using `Object#*` methods and `Array#reduce` as they basically discard all the type information about an object

cloudify: I don't like this method as it loses type information, in general I'm very much against using `Object#*` methods and `Array#reduce` as they basically discard all the type information about an object - in fact you can see it from the signature of the method, if the method returns only a subset of keys, ho can the output type be the same type of `obj`?

lib/models/notification.ts
gunzip: the method is still type safe btw. you cannot pick("[unknowProp"]) and you cannot omit required props (the returned object *must* be of type Notification). these are the same constraints you get whn using direct properties assignments.

lib/utils/types.ts
gunzip: which obj ? the type of the object that pick returns is responsibility of the caller. ie:

```
f(): SomeType {
  pick(...)   // <--- MUST return SomeType
}

f() {
  const a : SomeType = pick(...)   // <--- MUST return SomeType
}
```

cloudify: ok :)

lib/created_message_queue_handler.ts
gunzip: are you suggesting to pass an undefined value instead ? (ie. defaultAddress)

lib/created_message_queue_handler.ts
gunzip: done

cloudify:
My suggestion is to call this method only on defined values (possibly splitting this method in multiple methods covering the defined/not defined cases) and have the caller check the options

Inviato da iPhone

lib/controllers/messages.ts
```diff
@@ -276,21 +257,24 @@ async function getMessageNotificationStatuses(
     // collect the statuses of all channels
     const channelStatusesPromises = Object.keys(NotificationChannelEnum)
       .map(k => NotificationChannelEnum[k as NotificationChannelEnum])
-      .map(channel => ({
+      .map(async channel => ({
         channel,
-        status: getChannelStatus(channel)
+        status: await getChannelStatus(channel)
```
cloudify: I believe you don't want an await here, the map is synchronous and creates an array of promises, the await should happen just on Promise.all

cloudify: I believe you don't want an await here, the map is synchronous and creates an array of promises, the await should happen just on Promise.all

cloudify: see for instance [here](https://www.typescriptlang.org/play/#src=function%20f()%3A%20Promise%3Cstring%3E%20%7B%0A%20%20return%20Promise.resolve(%22%22)%3B%0A%7D%0A%0Aconst%20x%20%3D%20%5B%22a%22%5D.map(async%20e%20%3D%3E%20await%20f())%0A%0Aconst%20y%20%3D%20%5B%22a%22%5D.map(e%20%3D%3E%20f()))

gunzip: This was a patch to your PR as TS didn't compile complaining that there was an "unnecessary await on non promise value" (in the caller function, not here). This was because you cannot call Promise.all() on the results of this map() because the Promise is nested inside the returned object: you don't get an Array of Promises but an Array of { channel, Promise<status> }.

cloudify: ops yes sorry my mistake, yes code is good, thanks

cloudify:
could you please check why the coverage gets reduced so much? (-2.42%)

gunzip:
yes, I'll do a check (I guess because of the generated models in api/*)

gunzip:
I've added some tests

PR #149, opened 2018-02-14
refactor: upgrade to io-ts 1.x and fp-ts 1.x

gunzip:
- swapped t.Type<A, B> with t.Type<B, A>
- replaced all occurrences of t.validate(a, b) with b.decode(a)
- replaced calls to either.toOption() with fromEither()
- removed some superfluous conversions from eithers to options


lib/utils/__tests__/source_ip_check.test.ts
```diff
@@ -1,11 +1,10 @@
-import * as t from "io-ts";
-
 import { toAuthorizedCIDRs } from "../../models/service";
 import { ClientIp } from "../middlewares/client_ip_middleware";
 import { ResponseSuccessJson } from "../response";
 import { IPString } from "../strings";
 import { Tuple2 } from "../tuples";
 
+import { fromEither } from "fp-ts/lib/Option";
```
cloudify: I would import `Option` and call `Option.fromEither` below as `fromEither` alone is not really telling much, `Option.fromEither` will make it clear that we're converting an `Either` into an `Option`.

Btw I recall that `Either` had a `toOption` method, is it gone?

lib/utils/types.ts
```diff
@@ -111,17 +113,20 @@ export function strictInterfaceWithOptionals<
 >(
   required: R,
   optional: O,
-  name: string
-): t.Type<t.mixed, t.InterfaceOf<R> & t.PartialOf<O>> {
+  name?: string
+): t.Type<
+  { [K in keyof R]: t.TypeOf<R[K]> } & { [K in keyof O]?: t.TypeOf<O[K]> },
```
cloudify: perhaps this deserves a little explanation :)

also you could extract this complex types and give them some names to avoid inlining them here

lib/utils/types.ts
```diff
@@ -111,17 +113,20 @@ export function strictInterfaceWithOptionals<
...
   const loose = t.intersection([t.interface(required), t.partial(optional)]);
   const props = Object.assign({}, required, optional);
   return new t.Type(
-    name,
-    (v): v is t.InterfaceOf<R> & t.PartialOf<O> =>
-      loose.is(v) &&
-      Object.getOwnPropertyNames(v).every(k => props.hasOwnProperty(k)),
-    (s, c) =>
-      loose.validate(s, c).chain(o => {
+    name || `StrictInterfaceWithOptionals(${loose.name})`,
```
cloudify: ...just imagine the questions of somebody looking at this code for the first time :)

lib/utils/__tests__/source_ip_check.test.ts
gunzip: yes it is

PR #151, opened 2018-02-24
refactor: enum types

gunzip:
this is a better enum type than the actual one, as this one is statically checked
while actually you can create objects with a wrong enum key and it only fails at runtime

ref. https://github.com/gcanti/io-ts/issues/67#issuecomment-368068197

### DIFF
--- a/lib/utils/types.ts
+++ b/lib/utils/types.ts
@@ -1,4 +1,3 @@
-// tslint:disable:no-any
 import * as t from "io-ts";
 
 import { Set as SerializableSet } from "json-set-map";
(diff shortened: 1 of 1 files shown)


lib/utils/types.ts
```diff
@@ -12,27 +11,21 @@ export type Tagged<T, S extends t.mixed, A> = t.Type<A & T, S>;
...
-  const isE: (v: any) => boolean = v =>
-    typeof v === "string" && values.indexOf(v) >= 0;
-  return new t.Type<any, E>(
-    name,
-    (v): v is E => isE(v),
-    (v, c) => (isE(v) ? t.success(v) : t.failure(v, c)),
-    t.identity
+export const enumType = <E>(e: object, name: string): t.Type<E> => {
+  // get object values as keys
+  const keys = Object.keys(e).reduce(
```
cloudify: not sure I understand why this is needed, apparently you can just pass `e` to `t.keyof` instead of `keys`?

gunzip: only if you assume that keys and values are always the same (which it was not our assumption even in the current configuration, which is why we need a call to getObjectValues). probably it is more correct to work on values as keys can be integers (aka, key values depend on typescript internals while values are more trustworthy)

cloudify: yes indeed!

gunzip: that's the reason the method collects all the values and ignore keys (like the old getObjectValues)

cloudify: sorry I didn't understand at first, now I see what is going ok, perhaps it's worth splitting the method inside the reduce and/or add some comments

PR #158, opened 2018-03-03
[#150588117] WIP added time to live check to email notification handler

gunzip:
Between the main changes, note that with this PR [the whole Message is passed in the NotificationEvent](https://github.com/teamdigitale/digital-citizenship-functions/pull/158/files#diff-369872c88999ee654209f0771aaa8f25R129) (queue) while actually we pass only some fields:

- added two fields (createAt, timeToLive) to the MessageModel
- added a type (NotExpiredMessage) that checks that Date.now() < createdAt + timeToLive
- update notification status in case of error / expire
- refactor queue handlers to simplify error handling (ie. avoid try / catch blocks in index, pushed error logging at the end of the index method) and avoiding calling context.done (return a promise instead)

### DIFF
--- a/api/definitions.yaml
+++ b/api/definitions.yaml
@@ -69,7 +69,7 @@ NewMessage:
   type: object
   properties:
     time_to_live:
-      $ref: "#/TimeToLive"
+      $ref: "#/TimeToLiveSeconds"
     content:
       $ref: "#/MessageContent"
     default_addresses:
@@ -98,7 +98,7 @@ CreatedMessageWithContent:
     fiscal_code:
       $ref: "#/FiscalCode"
     time_to_live:
-      $ref: "#/TimeToLive"
+      $ref: "#/TimeToLiveSeconds"
     content:
       $ref: "#/MessageContent"
     sender_service_id:
@@ -114,7 +114,7 @@ CreatedMessageWithoutContent:
     fiscal_code:
       $ref: "#/FiscalCode"
     time_to_live:
-      $ref: "#/TimeToLive"
+      $ref: "#/TimeToLiveSeconds"
     sender_service_id:
       type: string
   required:
@@ -211,11 +211,11 @@ IsInboxEnabled:
   default: false
   description: True if the recipient of a message wants to store its content for
     later retrieval.
-TimeToLive:
+TimeToLiveSeconds:
   type: integer
   default: 3600
   minimum: 3600
-  maximum: 31536000
+  maximum: 604800
   description: This parameter specifies for how long (in seconds) the system will
     try to deliver the message to the channels configured by the user.
   example: 3600
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -110,12 +110,7 @@ gulp.task("unit:test", () => {
       // for some unknow reason jest tests won't exit on win32 machine
       // when run through gulp (running the tests directly through the jest CLI just works)
       // that's the reason of the `forceExit` flag
-      .pipe(
-        run(
-          "jest --coverage --runInBand" +
-            (process.platform === "win32" ? " --forceExit" : "")
-        )
-      )
+      .pipe(run("jest --coverage --runInBand"))
   );
 });
 
--- a/lib/controllers/messages.ts
+++ b/lib/controllers/messages.ts
@@ -16,7 +16,7 @@ import { IContext } from "azure-function-express";
 
 import { CreatedMessageWithoutContent } from "../api/definitions/CreatedMessageWithoutContent";
 import { FiscalCode } from "../api/definitions/FiscalCode";
-import { MessageContent } from "../api/definitions/MessageContent";
+
 import { MessageResponseWithContent } from "../api/definitions/MessageResponseWithContent";
 import { NewMessage as ApiNewMessage } from "../api/definitions/NewMessage";
 
@@ -60,6 +60,7 @@ import {
   ResponseErrorInternal,
   ResponseErrorNotFound,
   ResponseErrorQuery,
+  ResponseErrorValidation,
   ResponseSuccessJson,
   ResponseSuccessJsonIterator,
   ResponseSuccessRedirectToResource
@@ -85,6 +86,7 @@ import { BlobService } from "azure-storage";
 import {
   Message,
   MessageModel,
+  NewMessageWithContent,
   NewMessageWithoutContent,
   RetrievedMessage
 } from "../models/message";
@@ -96,7 +98,9 @@ import { fromEither, isNone, none, Option, some } from "fp-ts/lib/Option";
 import { CreatedMessageWithContent } from "../api/definitions/CreatedMessageWithContent";
 import { NotificationChannelEnum } from "../api/definitions/NotificationChannel";
 import { NotificationChannelStatusValueEnum } from "../api/definitions/NotificationChannelStatusValue";
+import { TimeToLive } from "../api/definitions/TimeToLive";
 import { NotificationStatusModel } from "../models/notification_status";
+import { readableReport } from "../utils/validation_reporters";
 
 /**
  * Input and output bindings for this function
@@ -211,6 +215,24 @@ type NotificationStatusHolder = Partial<
 >;
 
 /**
+ * Returns the status of a channel
+ */
+const getChannelStatus = async (
+  notificationStatusModel: NotificationStatusModel,
+  notificationId: NonEmptyString,
+  channel: NotificationChannelEnum
+) => {
+  const errorOrMaybeStatus = await notificationStatusModel.findOneNotificationStatusByNotificationChannel(
+    notificationId,
+    channel
+  );
+  return fromEither(errorOrMaybeStatus)
+    .chain(t.identity)
+    .map(o => o.status)
+    .toUndefined();
+};
+
+/**
  * Retrieve all notifications statuses (all channels) for a message.
  *
  * It makes one query to get the notification object associated
@@ -242,24 +264,16 @@ async function getMessageNotificationStatuses(
     }
     const notification = maybeNotification.value;
 
-    // returns the status of a channel
-    const getChannelStatus = async (channel: NotificationChannelEnum) => {
-      const errorOrMaybeStatus = await notificationStatusModel.findOneNotificationStatusByNotificationChannel(
-        notification.id,
-        channel
-      );
-      return fromEither(errorOrMaybeStatus)
-        .chain(t.identity)
-        .map(o => o.status)
-        .toUndefined();
-    };
-
     // collect the statuses of all channels
     const channelStatusesPromises = Object.keys(NotificationChannelEnum)
       .map(k => NotificationChannelEnum[k as NotificationChannelEnum])
       .map(async channel => ({
         channel,
-        status: await getChannelStatus(channel)
+        status: await getChannelStatus(
+          notificationStatusModel,
+          notification.id,
+          channel
+        )
       }));
     const channelStatuses = await Promise.all(channelStatusesPromises);
 
@@ -343,15 +357,25 @@ export function CreateMessageHandler(
       return ResponseErrorForbiddenNotAuthorizedForDefaultAddresses;
     }
 
+    const timeToLive = TimeToLive.decode(messagePayload.time_to_live);
+    if (isLeft(timeToLive)) {
+      return ResponseErrorValidation(
+        "Invalid message payload",
+        readableReport(timeToLive.value)
+      );
+    }
+
     // create a new message from the payload
     // this object contains only the message metadata, the content of the
-    // message is handled separately (see below)
+    // message is handled separately (see below).
     const newMessageWithoutContent: NewMessageWithoutContent = {
+      createdAt: new Date(),
       fiscalCode,
       id: generateObjectId(),
       kind: "INewMessageWithoutContent",
       senderServiceId: userService.serviceId,
-      senderUserId: auth.userId
+      senderUserId: auth.userId,
+      timeToLive: timeToLive.value
     };
 
     //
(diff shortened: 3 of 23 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#150588117](https://www.pivotaltracker.com/story/show/150588117): SA-8: Come SAS, quando invio un avviso, voglio poter impostare un tempo massimo di validità dell’avviso (time to live), per evitare che un avviso con problemi di invio venga inoltrato quando non è più valido.

<p align="right">
</p>

lib/controllers/messages.ts
```diff
@@ -210,6 +214,24 @@ type NotificationStatusHolder = Partial<
   Record<NotificationChannelEnum, NotificationChannelStatusValueEnum>
 >;
 
+/**
+ * Returns the status of a channel
+ */
+const getChannelStatus = async (
```
gunzip: moved outside the other function to avoid closures

lib/models/message.ts
```diff
@@ -141,6 +152,15 @@ export type RetrievedMessageWithoutContent = t.TypeOf<
   typeof RetrievedMessageWithoutContent
 >;
 
+export const NotExpiredMessage = t.refinement(
+  MessageBase,
```
gunzip: added

lib/models/notification_event.ts
```diff
@@ -13,8 +12,7 @@ import { CreatedMessageEventSenderMetadata } from "./created_message_sender_meta
  * have been configured for that notification.
  */
 export const NotificationEvent = t.interface({
-  messageContent: MessageContent,
-  messageId: NonEmptyString,
+  message: t.intersection([NewMessageWithContent, NotExpiredMessage]),
```
gunzip: notification event must carry a NotExpiredMessage

lib/utils/azure_queues.ts
```diff
@@ -35,77 +41,87 @@ export const getDelaySecForRetries = (
...
+ * @return              False if message is expired.
  */
-export function retryMessageEnqueue(
+export function updateMessageVisibilityTimeout<T extends IQueueMessage>(
   queueService: QueueService,
   queueName: string,
-  context: IContext
-): void {
-  const queueMessage = context.bindingData;
+  queueMessage: T
```
gunzip: I've refactored this function to not call context.done (just update the visibilityTimeout), which is a duty of the caller. In this way we can avoid to pass the whole context to this function.

gulpfile.js
```diff
@@ -110,12 +110,7 @@ gulp.task("unit:test", () => {
       // for some unknow reason jest tests won't exit on win32 machine
       // when run through gulp (running the tests directly through the jest CLI just works)
       // that's the reason of the `forceExit` flag
-      .pipe(
-        run(
-          "jest --coverage --runInBand" +
-            (process.platform === "win32" ? " --forceExit" : "")
-        )
-      )
+      .pipe(run("jest --coverage --runInBand"))
```
gunzip: we don't need this workaround anymore

lib/controllers/messages.ts
```diff
@@ -401,18 +425,21 @@ export function CreateMessageHandler(
     // emit created message event to the output queue
     //
 
-    const messageContent: MessageContent = {
-      markdown: messagePayload.content.markdown,
-      subject: messagePayload.content.subject
+    const newMessageWithContent: NewMessageWithContent = {
+      ...newMessageWithoutContent,
```
gunzip: This is a "big" change, as we pass the whole MessageWithContent here with this PR 
(actually we pass only some properties of the message object + messageContent)

lib/created_message_queue_handler.ts
```diff
@@ -149,29 +154,6 @@ function tryGetEmailNotification(
   });
 }
 
-/**
- * Save a notification object to the database.
- *
- * @param notificationModel notification model used to interact with database
```
gunzip: embededd in index(), we don't need this logic anymore

lib/created_message_queue_handler.ts
```diff
@@ -245,183 +218,187 @@ export async function handleMessage(
...
-      return context.done();
-    }
+    return left(TransientError("Cannot save notification to database"));
   }
 
-  const notification = errorOrNotification.value;
+  return right(errorOrNotification.value);
+}
 
+export function processGenericError(error: Error): void {
```
gunzip: logic here is we can 

- fail with a transient error -> trigger a retry
  -> check if the message has reached max number of retries -> update status
- fail with a permanent error -> stop message processing -> update status
- fail with some unexpected error (outside handleNotification) -> stop message processing -> update status
- success -> update status -> output bindings

lib/models/message.ts
```diff
@@ -16,22 +16,33 @@ import { MessageContent } from "../api/definitions/MessageContent";
...
+    // the identifier of the service of the sender
+    senderServiceId: t.string,
+
+    // the userId of the sender (this is opaque and depends on the API gateway)
+    senderUserId: NonEmptyString,
 
-  // the identifier of the service of the sender
-  senderServiceId: t.string,
+    // time to live in seconds
+    timeToLive: TimeToLive,
```
cloudify: can we call it `TimeToLiveSeconds` and `timeToLiveSeconds` ?

lib/models/notification_event.ts
cloudify: always better avoid naming things in a negative way (can cause confusion in double negation checks), perhaps  `ActiveMessage` ?

lib/utils/azure_queues.ts
```diff
@@ -3,11 +3,17 @@
...
 import { Option, some } from "fp-ts/lib/Option";
 
+export interface IQueueMessage extends QueueService.QueueMessageResult {
+  readonly id: string;
+  readonly nextVisibleTime: string;
+  readonly popReceipt: string;
+  readonly queueTrigger: string;
+}
+
 // Any delay must be less than 7 days (< 604800 seconds)
```
cloudify: would be nice to link to the relevant doc as this limit may change in the future

lib/controllers/messages.ts
```diff
@@ -343,15 +357,25 @@ export function CreateMessageHandler(
       return ResponseErrorForbiddenNotAuthorizedForDefaultAddresses;
     }
 
+    const timeToLive = TimeToLive.decode(messagePayload.time_to_live);
```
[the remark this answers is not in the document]
gunzip: That's because messagePayload carries an "ApiNewMessage" type where time_to_live is optional. As it's not guaranted that the middleware gets called before this point (ie. in tests), there is a double check here.

lib/controllers/messages.ts
gunzip: actually we pass only the message.id + message.content while now we pass the whole message to the email_notification_queue_handler (as it need to check ttl and createdAt):

https://github.com/teamdigitale/digital-citizenship-functions/pull/158/files#diff-369872c88999ee654209f0771aaa8f25L124

lib/controllers/messages.ts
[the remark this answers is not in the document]
gunzip: moreover typescript want that we assert that ttl is not undefined, so if you omit the decoding here you have to put one more if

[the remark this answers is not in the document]
cloudify: mmh no wait, the function signature guarantees to have a valid `ApiNewMessage` (aka `NewMessage`) as input (whether it comes from `MessagePayloadMiddleware` or something else is irrelevant) - and [NewMessage](https://github.com/teamdigitale/digital-citizenship-functions/blob/ae38da76c73362c082e799ca6d78b3d2ed3bad19/lib/api/definitions/NewMessage.ts#L27) is guaranteed to have `time_to_live` of type `TimeToLive | undefined` - so the right thing here is to just check whether `time_to_live` is defined or not, no need to validate, that *must* already have happened before.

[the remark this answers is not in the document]
gunzip: indeed you have to check here if ttl is undefined AND, if it is, use the default value (which is what decode() does)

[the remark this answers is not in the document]
cloudify: ahhh that's the problem then, this function doesn't want a `NewMessage`, this function wants a `NewMessage` where `undefined` values have been replaced by their defaults

the "problem" here is that `NewMessage` has `time_to_live` as optional when in fact is a required field (set to a default when not provided by the client)

it shouldn't be the concern of this function to decode payloads and replace undefined values with defaults, that must happen *before* by the middleware... and the input type should guarantee what this function expects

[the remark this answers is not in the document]
cloudify: Just keep this in mind: decoding and validation must happen only at the system boundary (e.g. with network or disk) - the responsibility of a middleware is to do this validation (decode `any` to a well defined type), ensuring that their output doesn't ever need to be validated again (indeed this is possible only when a type system can carry this information, like in Typescript).

So when you find yourself to have to do a validation inside the system, that is a smell of incorrect design of types.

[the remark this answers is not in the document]
cloudify: Just keep this in mind: decoding and validation must happen only at the system boundary (e.g. with network or disk) - the responsibility of a middleware is to do this validation (decode `any` to a well defined type), ensuring that their output doesn't ever need to be validated again (indeed this is possible only when a type system can carry this information, like in Typescript).

So when you find yourself to have to do a validation inside the system, that is a smell of incorrect design of types.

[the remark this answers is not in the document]
cloudify: Just keep this in mind: decoding and validation must happen only at the system boundary (e.g. with network or disk) - the responsibility of a middleware is to do this validation (decode `any` to a well defined type), ensuring that their output doesn't ever need to be validated again (indeed this is possible only when a type system can carry this information, like in Typescript).

So when you find yourself to have to do a validation inside the system, that is a smell of incorrect design of types.

lib/models/notification_event.ts
gunzip: renamed

lib/controllers/messages.ts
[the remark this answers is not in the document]
gunzip: done

lib/controllers/messages.ts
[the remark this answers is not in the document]
gunzip: refactored, see https://github.com/teamdigitale/digital-citizenship-functions/pull/158/files#diff-ca9342512f1db78ecb71b1a86cd82b77R113

Anyway the interface is rather clumsy because we have to keep this duality between input types (ttl optional, comes from the API call) and output types (ttl mandatory, to feed CreateMessageHandler) even if we know that calling ApiMessage.decode() will populate the ttl field (that's why we need another type ApiMessageWithDefaults)

lib/created_message_queue_handler.ts
```diff
@@ -245,183 +218,187 @@ export async function handleMessage(
...
-      winston.debug(
-        `CreatedMessageQueueHandler|validationError|${ReadableReporter.report(
-          validation
-        ).join("\n")}`
-      );
-      // we will never be able to recover from this, so don't trigger an error
-      // TODO: update message status (failed_at)
-      return context.done();
-    }
+  // it is an CreatedMessageEvent
```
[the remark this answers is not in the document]
gunzip: fixed

lib/emailnotifications_queue_handler.ts
```diff
@@ -114,6 +118,8 @@ const ContextWithBindings = t.interface({
 
 type ContextWithBindings = t.TypeOf<typeof ContextWithBindings> & IContext;
 
+type OutputBindings = undefined;
```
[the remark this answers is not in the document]
gunzip: that's right, fixed

lib/utils/azure_queues.ts
gunzip: actually this is an arbitrary value only set here. it *may* equal max(time to live), but not necessarily.

api/definitions.yaml
```diff
@@ -211,11 +211,11 @@ IsInboxEnabled:
   default: false
   description: True if the recipient of a message wants to store its content for
     later retrieval.
-TimeToLive:
+TimeToLiveSeconds:
   type: integer
   default: 3600
   minimum: 3600
-  maximum: 31536000
+  maximum: 604800
```
cloudify: hehe told you :D even the most experienced programmers get caught by these kind of bugs, keeping track of unit of measure is hard, esp. re time

gunzip:
@cloudify 

I've refactored the two queue handlers to they don't call context.done anymore (they just return a promise with bindings)

see here:
https://github.com/teamdigitale/digital-citizenship-functions/pull/158/files#diff-1f1ba005c775c0ea6e2bd5589e5dc6cbR365

and here:
https://github.com/teamdigitale/digital-citizenship-functions/pull/158/files#diff-642e2f59547d5534a7b9902770a97a6fR335

lib/created_message_queue_handler.ts
cloudify: ditto

lib/created_message_queue_handler.ts
```diff
@@ -312,8 +332,10 @@ export function processSuccess(
 /**
  * Handler that gets triggered on incoming event.
  */
-export function index(context: ContextWithBindings): void {
-  // redirect winston logs to Azure Functions log
+export async function index(
+  context: ContextWithBindings
+): Promise<OutputBindings | Error | void> {
+  const stopProcessing = undefined;
```
cloudify: I suggest you return `undefined` and add a comment saying "Returns undefined to stop processing"

gunzip: we can do it but using a variable instead of 'undefined' lets you track where the process exits (while 'undefined' values may show up in some other unrelated places)

PR #159, opened 2018-03-04
[#155679212] fix superagent mock and avoid jest tests to stuck

gunzip:
It looks like superagent-mock does not work with promises as it expects the metehod end() to be called.
As we don't want to fallback to old style callbacks, it's necessary to provide a custom superagent mock implementation. An alternative is to fake NodeJS http requests at a lower level with something like https://github.com/node-nock/nock

### DIFF
--- a/__mocks__/superagent.ts
+++ b/__mocks__/superagent.ts
@@ -1,0 +1,44 @@
+/**
+ * Implements a superagent mock to use inside jest tests.
+ * Avoids real calls to some HTTP backend.
+ *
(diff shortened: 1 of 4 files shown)


digitalcitizenship:
## Affected stories

  * ⚙️ [#155679212](https://www.pivotaltracker.com/story/show/155679212): Investigare perché i test su circleci non terminano in un tempo ragionevole

<p align="right">
</p>

cloudify:
Awesome, back to reasonable execution time 🚀

PR #161, opened 2018-03-15
Use a standard parameter from the standard definitions.

ioggstream:
## I expect

that common arguments should be in the definitions and/or referenced via $ref


api/public_api_v1.yaml
```diff
@@ -241,12 +241,7 @@ definitions:
 responses: {}
 parameters:
   PaginationRequest:
-    name: cursor
-    in: query
-    type: string
-    minimum: 1
-    description: An opaque identifier that points to the next item in the collection.
-    x-example: 01BX9NSMKVXXS5PSP2FATZM123
+    $ref: 'https://raw.githubusercontent.com/ioggstream/openapi-samples/master/openapi-v2/schema/components.yaml#/x-components/parameters/cursor'
```
cloudify: instead of `master` you should ref a specific commit, or better a specific tag

cloudify:
@gunzip what do you think? is this going to break our model generation script?

api/public_api_v1.yaml
cloudify: also, you should reference an official org (e.g. `teamdigitale` or `italia`), not a personal account

gunzip: AFAIK think this PR is a just a "memento" :) I'd integrate the new interoperability specs as a whole 
once they're stable.

ioggstream:
Close and reopen with a comprehensive patch.

PR #162, opened 2018-03-16
[#150737636] add message status

gunzip:
- added versioned message status model
- update message status in the message queue handler
- the getmessage controller now returns message status aside notification status

Shape of a message:

```
{
    "message": {
        "content": {
            "markdown": "...",
            "subject": "..."
        },
        "fiscal_code": "ZERTHB99A45Y149X",
        "id": "01C8RHV61QBX82GSWYQ1VFNBN5",
        "sender_service_id": "5a25"
    },
    "notification": {
        "email": "SENT_TO_CHANNEL"
    },
    "status": "PROCESSED" <------
}

```

### DIFF
--- a/api/definitions.yaml
+++ b/api/definitions.yaml
@@ -37,8 +37,14 @@ NotificationChannel:
   example: EMAIL
 NotificationChannelStatusValue:
   type: string
+    The status of a notification (one for each channel).
+    "SENT_TO_CHANNEL": the notification was succesfully sent to the channel (ie. email or push notification)
+    "THROTTLED": a temporary failure caused a retry during the notification processing;
+                 any notification associated with this channel will be delayed until the message expires
+    "EXPIRED": the message expired before the notification could be sent;
+               this means that the maximum message time to live was reached so no notification will be sent for this channel
+    "FAILED": a permanent failure caused the process to exit with an error, no notification for this channel will be sent for this message
   x-extensible-enum:
-    - QUEUED
     - SENT_TO_CHANNEL
     - THROTTLED
     - EXPIRED
--- a/api/definitions.yaml
+++ b/api/definitions.yaml
@@ -37,6 +37,7 @@ NotificationChannel:
   example: EMAIL
 NotificationChannelStatusValue:
   type: string
+  description: |-
     The status of a notification (one for each channel).
     "SENT_TO_CHANNEL": the notification was succesfully sent to the channel (ie. email or push notification)
     "THROTTLED": a temporary failure caused a retry during the notification processing;
--- a/api/definitions.yaml
+++ b/api/definitions.yaml
@@ -50,12 +50,14 @@ NotificationChannelStatus:
       $ref: "#/NotificationChannel"
     status:
       $ref: "#/NotificationChannelStatusValue"
-    updateAt:
+    updated_at:
       $ref: "#/Timestamp"
+    version:
+      type: integer
   required:
     - channel
     - status
-    - updateAt
+    - updated_at
 MessageContent:
   type: object
   properties:
@@ -78,18 +80,31 @@ NewMessage:
     - content
 MessageStatusValue:
   type: string
+  description: |-
+    The processing status of a message.
+    "ACCEPTED": the message has been accepted and will be processed so we can try to notify the recipient
+    "THROTTLED": a temporary failure caused a retry during the message processing;
+                 any notification associated with this message will be delayed for a maximum of 7 days
+    "FAILED": a permanent failure caused the process to exit with an error, no notification will be sent for this message
+    "PROCESSED": the message was processed succesfully, we'll try to send a notification for each of the selected channels
   x-extensible-enum:
-    - PROCESSING
     - ACCEPTED
+    - THROTTLED
     - FAILED
+    - PROCESSED
   example: ACCEPTED
 MessageStatus:
   type: object
   properties:
     status:
       $ref: "#/MessageStatusValue"
-    updateAt:
+    updated_at:
       $ref: "#/Timestamp"
+    version:
+      type: integer
+  required:
+    - status
+    - updated_at
 CreatedMessageWithContent:
   type: object
   properties:
@@ -132,6 +147,8 @@ MessageResponseWithContent:
       $ref: "#/CreatedMessageWithContent"
     notification:
       $ref: "#/MessageResponseNotificationStatus"
+    status:
+      $ref: "#/MessageStatusValue"
   required:
     - message
 MessageResponseWithoutContent:
@@ -141,6 +158,8 @@ MessageResponseWithoutContent:
       $ref: "#/CreatedMessageWithoutContent"
     notification:
       $ref: "#/MessageResponseNotificationStatus"
+    status:
+      $ref: "#/MessageStatusValue"
   required:
     - message
 FiscalCode:
--- a/api/definitions.yaml
+++ b/api/definitions.yaml
@@ -41,10 +41,10 @@ NotificationChannelStatusValue:
     The status of a notification (one for each channel).
     "SENT_TO_CHANNEL": the notification was succesfully sent to the channel (ie. email or push notification)
     "THROTTLED": a temporary failure caused a retry during the notification processing;
-                 any notification associated with this channel will be delayed until the message expires
+      the notification associated with this channel will be delayed until the message expires
     "EXPIRED": the message expired before the notification could be sent;
-               this means that the maximum message time to live was reached so no notification will be sent for this channel
-    "FAILED": a permanent failure caused the process to exit with an error, no notification for this channel will be sent for this message
+      this means that the maximum message time to live was reached so no notification will be sent to this channel
+    "FAILED": a permanent failure caused the process to exit with an error, no notification will be sent to this channel
   x-extensible-enum:
     - SENT_TO_CHANNEL
     - THROTTLED
@@ -92,7 +92,7 @@ MessageStatusValue:
     The processing status of a message.
     "ACCEPTED": the message has been accepted and will be processed so we can try to notify the recipient
     "THROTTLED": a temporary failure caused a retry during the message processing;
-                 any notification associated with this message will be delayed for a maximum of 7 days
+      any notification associated with this message will be delayed for a maximum of 7 days
     "FAILED": a permanent failure caused the process to exit with an error, no notification will be sent for this message
     "PROCESSED": the message was processed succesfully, we'll try to send a notification for each of the selected channels
   x-extensible-enum:
--- a/api/definitions.yaml
+++ b/api/definitions.yaml
@@ -39,18 +39,18 @@ NotificationChannelStatusValue:
   type: string
   description: |-
     The status of a notification (one for each channel).
-    "SENT_TO_CHANNEL": the notification was succesfully sent to the channel (ie. email or push notification)
+    "SENT": the notification was succesfully sent to the channel (ie. email or push notification)
     "THROTTLED": a temporary failure caused a retry during the notification processing;
-      the notification associated with this channel will be delayed until the message expires
+      the notification associated with this channel will be delayed for a maximum of 7 days or until the message expires
     "EXPIRED": the message expired before the notification could be sent;
-      this means that the maximum message time to live was reached so no notification will be sent to this channel
+      this means that the maximum message time to live was reached; no notification will be sent to this channel
     "FAILED": a permanent failure caused the process to exit with an error, no notification will be sent to this channel
   x-extensible-enum:
-    - SENT_TO_CHANNEL
+    - SENT
     - THROTTLED
     - EXPIRED
     - FAILED
-  example: SENT_TO_CHANNEL
+  example: SENT
 NotificationChannelStatus:
   type: object
   properties:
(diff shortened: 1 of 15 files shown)


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#150737636](https://www.pivotaltracker.com/story/show/150737636): Come SAS voglio sapere se CIT ha letto il messaggio che gli ho inviato.

<p align="right">
</p>

api/definitions.yaml
```diff
@@ -79,17 +81,23 @@ NewMessage:
 MessageStatusValue:
   type: string
   x-extensible-enum:
-    - PROCESSING
     - ACCEPTED
     - FAILED
+    - PROCESSED
```
cloudify: we need to document a state diagram somewhere that describes the initial and final states and their meanings

lib/controllers/messages.ts
```diff
@@ -606,15 +633,31 @@ export function GetMessageHandler(
...
+    const maybeMessageStatus = errorOrMaybeMessageStatus.value;
+
+    const returnedMessage: MessageResponseWithContent = {
       message,
-      notification: notificationStatuses.toUndefined()
+      notification: notificationStatuses.toUndefined(),
+      // we do not return the status date-time
+      status: maybeMessageStatus
+        .map(messageStatus => messageStatus.status)
+        .toUndefined()
```
cloudify: getOrElse(ACCEPTED)

lib/created_message_queue_handler.ts
```diff
@@ -258,8 +269,11 @@ export async function handleMessage(
   return right(errorOrNotification.value);
 }
 
-export function processGenericError(error: Error): void {
-  // TODO: update message status (FAILED)
+export async function processGenericError(
+  messageStatusUpdater: MessageStatusUpdater,
+  error: Error
+): Promise<void> {
+  await messageStatusUpdater(MessageStatusValueEnum.FAILED);
```
cloudify: this should fail the whole request and trigger a retry

lib/created_message_queue_handler.ts
```diff
@@ -281,16 +296,17 @@ export async function processRuntimeError(
     );
     return left(shouldTriggerARetry);
   } else {
-    // TODO: update message status (FAILED)
     winston.error(
       `CreatedMessageQueueHandler|Permanent error|${error.message}`
     );
+    await messageStatusUpdater(MessageStatusValueEnum.FAILED);
```
cloudify: check error and fail function

lib/created_message_queue_handler.ts
```diff
@@ -321,9 +337,10 @@ export async function processSuccess(
     emailNotification.message.id,
     emailNotification.notificationId
   );
-  await notificationStatusUpdater(NotificationChannelStatusValueEnum.QUEUED);
 
-  // TODO: update message status (ACCEPTED)
+  // status updates are best effort operations, no error checking here
```
cloudify: check errors

lib/controllers/messages.ts
```diff
@@ -606,15 +615,31 @@ export function GetMessageHandler(
         }|${errorOrNotificationStatuses.value.message}`
       );
     }
-
     const notificationStatuses = errorOrNotificationStatuses.value;
 
-    const messageStatus: MessageResponseWithContent = {
+    const errorOrMaybeMessageStatus = await messageStatusModel.findOneByMessageId(
```
gunzip: returns message status from the GetMessage API call

lib/created_message_queue_handler.ts
```diff
@@ -101,6 +103,33 @@ const messageContainerName = getRequiredStringEnv("MESSAGE_CONTAINER_NAME");
 const storageConnectionString = getRequiredStringEnv("QueueStorageConnection");
 const queueConnectionString = getRequiredStringEnv("QueueStorageConnection");
 
+const documentClient = new DocumentDBClient(cosmosDbUri, {
```
gunzip: I've moved the inizialization code for models outside index() as 
1. this is more coherent on what we do in the api handler (public_api_v1.ts)
2. this will eventually fail at deploy time (not a runtime)
3. it's impossible to handle this kind of errors with retries (as we need the models and services to be valid)

lib/created_message_queue_handler.ts
```diff
@@ -174,16 +203,16 @@ function tryGetEmailNotification(
  *          a PermanentError in case of unrecoverable error
  */
 export async function handleMessage(
-  profileModel: ProfileModel,
-  messageModel: MessageModel,
-  notificationModel: NotificationModel,
-  blobService: BlobService,
+  lProfileModel: ProfileModel,
```
gunzip: renamed local variables to not shadow the global ones

lib/created_message_queue_handler.ts
```diff
@@ -258,77 +287,58 @@ export async function handleMessage(
   return right(errorOrNotification.value);
 }
 
-export function processGenericError(error: Error): void {
```
gunzip: removed

lib/created_message_queue_handler.ts
```diff
@@ -258,77 +287,58 @@ export async function handleMessage(
...
 ): Promise<Either<boolean, Option<OutputBindings>>> {
   if (isTransient(error)) {
     winston.warn(`CreatedMessageQueueHandler|Transient error|${error.message}`);
     const shouldTriggerARetry = await updateMessageVisibilityTimeout(
-      queueService,
+      lQueueService,
       MESSAGE_QUEUE_NAME,
       queueMessage
     );
+    const errorOrUpdatedMessageStatus = await messageStatusUpdater(
```
gunzip: check errors during status updates

lib/created_message_queue_handler.ts
```diff
@@ -258,77 +287,58 @@ export async function handleMessage(
...
-    // return no bindings so message processing stops here
-    return right(none);
+    const errorOrUpdatedMessageStatus = await messageStatusUpdater(
+      MessageStatusValueEnum.FAILED
+    );
+    if (isLeft(errorOrUpdatedMessageStatus)) {
+      // if we get an error updating the message status
+      // we retry the whole procedure (it will log the error anyway)
+      const transientError = errorOrUpdatedMessageStatus.value;
+      return await processRuntimeError(
```
gunzip: this is a recursive call which "translates" a permanent error to a transient error (the status update failed, we trigger a retry)

lib/created_message_queue_handler.ts
```diff
@@ -258,77 +287,58 @@ export async function handleMessage(
...
+        transientError
+      );
+    } else {
+      // return no bindings so message processing stops here
+      return right(none);
+    }
   }
 }
 
-export async function processSuccess(
```
gunzip: removed, embedded in the index, see below

lib/created_message_queue_handler.ts
```diff
@@ -358,53 +368,26 @@ export async function index(
...
-    const senderMetadata = createdMessageEvent.senderMetadata;
-
-    winston.info(
-      `CreatedMessageQueueHandler|A new message was created|${
-        newMessageWithContent.id
-      }|${newMessageWithContent.fiscalCode}`
-    );
-
-    // setup required models
-    const documentClient = new DocumentDBClient(cosmosDbUri, {
```
gunzip: moved to the outside global scope

lib/created_message_queue_handler.ts
```diff
@@ -415,32 +398,50 @@ export async function index(
...
+
+    winston.debug(
+      `CreatedMessageQueueHandler|succeeded|message=${
+        emailNotification.notificationId
+      }|message=${emailNotification.message.id}`
     );
+
+    return { emailNotification };
+    //
   } catch (error) {
```
gunzip: we cannot just return eithers and map here as at some point we 
*must* catch unknow exceptions which may occur inside library code

lib/created_message_queue_handler.ts
```diff
@@ -415,32 +398,50 @@ export async function index(
...
-    if (isTransient(error)) {
-      throw error;
+    const shouldTriggerARetry = await processRuntimeError(
+      queueService,
+      messageStatusUpdater,
+      context.bindingData,
+      RuntimeErrorOf(error)
```
gunzip: we don't know what we are catching here

lib/emailnotifications_queue_handler.ts
```diff
@@ -87,6 +88,23 @@ const notificationStatusCollectionUrl = documentDbUtils.getCollectionUri(
 export const EMAIL_NOTIFICATION_QUEUE_NAME = "emailnotifications";
 const queueConnectionString = getRequiredStringEnv("QueueStorageConnection");
 
+const documentClient = new DocumentDBClient(cosmosDbUri, {
```
gunzip: moved to the module scope

lib/emailnotifications_queue_handler.ts
```diff
@@ -307,58 +325,56 @@ export async function handleNotification(
...
       `EmailNotificationQueueHandler|Transient error|${error.message}`
     );
     const shouldTriggerARetry = await updateMessageVisibilityTimeout(
-      queueService,
+      lQueueService,
       EMAIL_NOTIFICATION_QUEUE_NAME,
       queueMessage
     );
+    const errorOrNotificationStatus = await notificationStatusUpdater(
+      NotificationChannelStatusValueEnum.THROTTLED
```
gunzip: in case of retries, update the notification status to "throttled"

lib/emailnotifications_queue_handler.ts
```diff
@@ -307,58 +325,56 @@ export async function handleNotification(
...
-    // return no bindings so message processing stops here
-    return right(none);
+    const errorOrNotificationStatus = await notificationStatusUpdater(
+      NotificationChannelStatusValueEnum.FAILED
+    );
+    if (isLeft(errorOrNotificationStatus)) {
+      // if we get an error updating the notification status
+      // we retry the whole procedure (it will log the error anyway)
+      const transientError = errorOrNotificationStatus.value;
+      return await processRuntimeError(
```
gunzip: recursive call (in case the status update fail we trigger a retry)

lib/emailnotifications_queue_handler.ts
```diff
@@ -307,58 +325,56 @@ export async function handleNotification(
...
+        transientError
+      );
+    } else {
+      // return no bindings so message processing stops here
+      return right(none);
+    }
   }
 }
 
-export async function processSuccess(
```
gunzip: removed, handled in the index

lib/models/message_status.ts
```diff
@@ -0,0 +1,170 @@
+import * as t from "io-ts";
```
gunzip: this is the notification status model

lib/models/notification_status.ts
```diff
@@ -47,7 +47,7 @@ export const NotificationStatus = t.interface({
   // As we have one NotificationStatus for each channel
   // of a Notification => statusId = notificationId + channelName
   statusId: NotificationStatusId,
-  updateAt: Timestamp
+  updatedAt: Timestamp
```
gunzip: there was a typo

lib/controllers/messages.ts
```diff
@@ -606,15 +615,31 @@ export function GetMessageHandler(
...
+    const maybeMessageStatus = errorOrMaybeMessageStatus.value;
+
+    const returnedMessage: MessageResponseWithContent = {
       message,
-      notification: notificationStatuses.toUndefined()
+      notification: notificationStatuses.toUndefined(),
+      // we do not return the status date-time
+      status: maybeMessageStatus
+        .map(messageStatus => messageStatus.status)
+        .getOrElse(MessageStatusValueEnum.ACCEPTED)
```
cloudify: a brief explanation of this behavior would be useful here (e.g. when the message has been stored but a `MessageStatus` does not exist, the message is considered to be in the `ACCEPTED` state)

lib/controllers/messages.ts
```diff
@@ -606,15 +615,31 @@ export function GetMessageHandler(
...
     };
 
-    return ResponseSuccessJson(messageStatus);
+    return ResponseSuccessJson(withoutUndefinedValues(returnedMessage));
```
cloudify: split this in two steps and explain why `withoutUndefinedValues` is needed

lib/created_message_queue_handler.ts
cloudify: I don't understand reason n.2, what do you mean? (also put these reasons as comments in code, e.g. we're initializing the db client here because...)

lib/created_message_queue_handler.ts
```diff
@@ -101,6 +103,33 @@ const messageContainerName = getRequiredStringEnv("MESSAGE_CONTAINER_NAME");
...
+const documentClient = new DocumentDBClient(cosmosDbUri, {
+  masterKey: cosmosDbKey
+});
+
+const messageStatusModel = new MessageStatusModel(
+  documentClient,
+  messageStatusCollectionUrl
+);
+
+// needed for retries
```
cloudify: "We cannot use Functions bindings for retries because ... so we have to instantiate a client for the Queue storage"

lib/created_message_queue_handler.ts
cloudify: this is a code smell, this method should be probably moved to a separate file

lib/created_message_queue_handler.ts
gunzip: reason 2 is wrong (I thought that the function runtime would require the module at startup but it doesn't)

lib/created_message_queue_handler.ts
gunzip: I agree, there's a pivotal task with "split controllers in multiple files", probably this applies to queue handlers as well (but I wouln't do this with this PR).

Moreover, now, the error handling of the two queue handlers (processRuntimeError and the try / catch block) is very similar so probably we can do a refactor to abstract the shared code away
(ie. a function tryProcess(queueHandler) that do the error handling so we can avoid to throw in the index method).

lib/controllers/messages.ts
gunzip: I've removed it (it's not really needed, it was just to avoid return { email: undefined } for the notification statuses)

PR #164, opened 2018-03-21
[#156153225] remove documentdb update

gunzip:
Note that at the moment the update method is unused


digitalcitizenship:
## Affected stories

  * 🐞 [#156153225](https://www.pivotaltracker.com/story/show/156153225): Fix race condition in documentdb update

<p align="right">
</p>

lib/utils/documentdb.ts
```diff
@@ -469,6 +469,10 @@ export function replaceDocument<T>(
       documentUri.uri,
       document,
       {
+        accessCondition: {
+          condition: document._etag,
```
cloudify: I don't understand what's happening here, `document` is the *new* document, how can its `etag` match the `etag` of the existing document?

lib/utils/documentdb.ts
```diff
@@ -462,15 +462,23 @@ export function replaceDocument<T>(
   client: DocumentDb.DocumentClient,
   documentUri: IDocumentDbDocumentUri,
   document: T & DocumentDb.NewDocument,
-  partitionKey: string
+  partitionKey: string,
+  etag?: string
 ): Promise<Either<DocumentDb.QueryError, T & DocumentDb.RetrievedDocument>> {
   return new Promise(resolve => {
+    const requestOptions = fromNullable(etag)
```
gunzip: for some unknow reason the _etag field is marked as optional in the RetrievedDocument typings

lib/utils/documentdb.ts
```diff
@@ -462,19 +462,23 @@ export function replaceDocument<T>(
   client: DocumentDb.DocumentClient,
   documentUri: IDocumentDbDocumentUri,
   document: T & DocumentDb.NewDocument,
-  partitionKey: string
+  partitionKey: string,
+  etag?: string
```
cloudify: this should not be optional, as you're replacing an existing document, you should always provide the `etag` of the existing document - I suggest to throw an error on the caller site in case the `etag` is not defined

gunzip: To do that we must patch the typings of RetrievedDocument where etag is optional:

```
/** Represents the common meta data for all DocumentDB objects. */
export interface AbstractMeta extends UniqueId {
    /** The self link. */
    _self: string;

    /** The time the object was created. */
    _ts: number;

    _rid?: string;
    _etag?: string;

    _attachments?: string;
}
```

This because the etag value comes from some RetrievedDocument returned by the find method: 

https://github.com/teamdigitale/digital-citizenship-functions/blob/4836559cf4370a5b2bdb71ba5d8499d1829add2a/lib/utils/documentdb_model.ts#L148

cloudify: I was thinking about something like:

```ts
const doc = findDoc(...);
const currentETag = doc._etag;
if(currentETag === undefined) {
  throw new Error("OMG");
}
client.update(newDoc, doc._etag);

```

gunzip: anoter option is to get rid of this method entirely as everything should be immutable

cloudify: I like this option

gunzip: done

PR #165, opened 2018-03-21
Use standard parameters for pagination requests.

ioggstream:
## What does this PR do

 - add `components.yaml` from https://github.com/teamdigitale/api-openapi-samples
 - reference the `cursor` parameter from `components.yaml`
 - provide common throttling headers and error statuses

## Goal

  - share common parameters/structures with different API project as a baseline for implementing generic API features.
  - get feedback from different API teams about parameter standardization


gunzip:
1. actually we don't implement any pagination so it will be ok to merge the pagination query parameters which lgtm
2. about the "Error" definition, we use ProblemJson as described in [zalando api](https://github.com/zalando/problem) so to merge that part we must refactor our error responses
3. about the rate limiting http headers, we don't have any as there's no way to expose quota in the azure api management gateway

api/public_api_v1.yaml
```diff
@@ -241,12 +241,7 @@ definitions:
 responses: {}
 parameters:
   PaginationRequest:
-    name: cursor
-    in: query
-    type: string
-    minimum: 1
-    description: An opaque identifier that points to the next item in the collection.
-    x-example: 01BX9NSMKVXXS5PSP2FATZM123
+    $ref: 'components.yaml#/x-components/parameters/cursor'
```
ioggstream: @gunzip Is the `cursor` still implemented or wip?

api/components.yaml
```diff
@@ -0,0 +1,81 @@
+swagger: '2.0'
+definitions:
+  Error:
```
ioggstream: @gunzip Let me check https://github.com/zalando/problem. That could be the new standard ;)

api/components.yaml
```diff
@@ -0,0 +1,81 @@
...
+          schema:
+            $ref: '#/definitions/Error'
+      description: Service Unavailable
+    default:
+      content:
+        application/json:
+          schema:
+            $ref: '#/definitions/Error'
+      description: Unexpected error
+  headers:
```
ioggstream: @gunzip
While throttling it's not important at development time, it MUST be supported once they are exposed (Eg. via the api gateway). So it's fine for now, but put a note on the production environment ;)

gunzip: we do use throttling but there are no exposed headers with quotas (and with the current api gateway we have no chance to do that =)

api/public_api_v1.yaml
gunzip: actually we don't use pagination at all but it's planned

gunzip:
to recap, we cannot merge this PR actually because:

1. we use ProblemJSON instead of the Error schema proposed
1. our tool that generates code from specs does not support the x-components directive
1. we don't expose any http header with api limits / quota

One more thing which is not totally clear to me:

```
  responses:
    400BadRequest:
```

shouldn't be

```
  responses:
   ' 400':
```

as described in the reference: https://swagger.io/docs/specification/describing-responses/ ?

cloudify:
closing as this PR looks problematic

PR #167, opened 2018-03-24
[#156125470] refactor of request middleware 

gunzip:
this is a proposal to refactor wrapRequestMiddleware method. 

while the internal logic is now untyped, this method is not exposed to the calles so type safety is preserved in the calling code.

This PR embeds some mods from #168 otherwise tests would fail.

### DIFF
--- a/lib/utils/request_middleware.ts
+++ b/lib/utils/request_middleware.ts
@@ -194,19 +194,15 @@ function withRequestMiddlewaresAr(
     // by the handler or one of the types returned by any of the middlewares (i.e., when
     // a middleware returns an error response).
     return request => {
-      // make an array of middleware tasks
-      // to defer their execution
-      const tasks = mws.reduce(
-        (prev, curr) => {
-          return curr !== undefined
-            ? prev.concat([fromTask(() => curr(request))])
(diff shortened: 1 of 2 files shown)


lib/utils/request_middleware.ts
```diff
@@ -172,120 +179,39 @@ export function withRequestMiddlewares<
...
-              handler(r1.value).then(resolve, reject);
-            }
-          }, reject);
-        }
+  return withRequestMiddlewaresAr([v1, v2, v3, v4, v5, v6]);
+}
+
+/* tslint:disable:no-any */
+/* tslint:disable:readonly-array */
+function withRequestMiddlewaresAr(
```
gunzip: private untyped internal implementation, not exposed to the caller

digitalcitizenship:
## Affected stories

  * ⚙️ [#156125470](https://www.pivotaltracker.com/story/show/156125470): Rifattorizzare la gestione degli errori e del retry negli handler delle code

<p align="right">
</p>

gunzip:
closing as I realized that this runs the middleware in parallel (which is not always expected)

PR #171, opened 2018-03-28
[#156125470] refactor queue handlers

gunzip:
The aim of this PR is to abstract away the error handling in queue handlers.

There are different commit for tests and handlers (to make review easier)

The logic that causes the process to stop / retry in case of permanent / transient errors is moved in a dedicated reusable function in azure_queues.ts.


lib/created_message_queue_handler.ts
```diff
@@ -289,59 +280,18 @@ export async function handleMessage(
...
-        messageStatusUpdater,
-        queueMessage,
-        transientError
-      );
-    } else {
-      // message processing stops here
-      return false;
-    }
-  }
+  const emailNotification: NotificationEvent = {
```
gunzip: this logic was in the index() method, moved here.

lib/created_message_queue_handler.ts
```diff
@@ -289,59 +280,18 @@ export async function handleMessage(
...
 
-  return right(errorOrNotification.value);
-}
+  const notification = errorOrNotification.value;
 
-/**
- * Returns true in case the caller must
- * trigger a retry
- */
-export async function processRuntimeError(
```
gunzip: deleted. now it is implemented as a generic method in azure_queue.ts

lib/created_message_queue_handler.ts
```diff
@@ -392,61 +339,41 @@ export async function index(
...
-    // try to update the message status in the database
-    const errorOrUpdatedMessageStatus = await messageStatusUpdater(
-      MessageStatusValueEnum.PROCESSED
-    );
-    if (isLeft(errorOrUpdatedMessageStatus)) {
-      // this will trigger a retry as it's considered a transient error
-      throw errorOrUpdatedMessageStatus.value;
-    }
-
-    // everything went well, make output bindings
```
gunzip: moved in handleMessage()

lib/created_message_queue_handler.ts
```diff
@@ -392,61 +339,41 @@ export async function index(
...
+    .catch(error =>
+      handleQueueProcessingFailure(
+        queueService,
+        context.bindingData,
+        MESSAGE_QUEUE_NAME,
+        // execute in case of transient errors
+        () => messageStatusUpdater(MessageStatusValueEnum.THROTTLED),
+        // execute in case of permanent errors
+        () => messageStatusUpdater(MessageStatusValueEnum.FAILED),
+        error
```
gunzip: this method throws an exception to trigger a retry

lib/emailnotifications_queue_handler.ts
```diff
@@ -342,56 +350,6 @@ export async function handleNotification(
   return right(emailNotificationEvent);
 }
 
-export async function processRuntimeError(
```
gunzip: deleted, see above comments.

lib/utils/azure_queues.ts
```diff
@@ -107,10 +114,11 @@ export function updateMessageVisibilityTimeout<T extends IQueueMessage>(
           visibilityTimeoutSec,
           err => {
             winston.info(
-              err.message ||
-                `Updated visibilityTimeout|retry=${numberOfRetries}|timeout=${visibilityTimeoutSec}|queueMessageId=${
-                  queueMessage.id
-                }`
+              err
+                ? err.message
```
gunzip: fixed a nasty bug (err can be undefined)

lib/utils/azure_queues.ts
```diff
@@ -127,3 +135,65 @@ export function updateMessageVisibilityTimeout<T extends IQueueMessage>(
       });
   });
 }
+
+/**
+ * Returns false when we need to trigger a retry.
+ */
+export async function handleQueueProcessingFailure(
```
gunzip: this is the generalized error handling logic for queue processors.

digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * ⚙️ [#156125470](https://www.pivotaltracker.com/story/show/156125470): Rifattorizzare la gestione degli errori e del retry negli handler delle code

<p align="right">
</p>

lib/utils/azure_queues.ts
```diff
@@ -127,3 +135,68 @@ export function updateMessageVisibilityTimeout<T extends IQueueMessage>(
...
+  queueService: QueueService,
+  queueMessage: IQueueMessage,
+  queueName: string,
+  onTransientError: () => Promise<Either<RuntimeError, {}>>,
+  onPermanentError: () => Promise<Either<RuntimeError, {}>>,
+  error: Error | RuntimeError
```
cloudify: this is quite confusing, this function accepts an error of type `Error` but it handles it as a `PermanentError`, that is a subtype of  `RuntimeError` - I suggest to just accept a `RuntimeError` here

lib/utils/azure_queues.ts
```diff
@@ -127,3 +135,68 @@ export function updateMessageVisibilityTimeout<T extends IQueueMessage>(
...
+): Promise<void> {
+  const runtimeError = RuntimeErrorOf(error);
```
cloudify: by looking at this code it's really hard to guess what `RuntimeErrorOf` is supposed to do

lib/utils/errors.ts
```diff
@@ -50,8 +54,16 @@ export const of = (error: any): RuntimeError =>
...
+  | PermanentError
+  | UnknowError
+  | ExpiredError;
 
 // tslint:disable-next-line:no-any
 export const isTransient = (error: any): error is TransientError =>
   error.kind && error.kind === ErrorTypes.TransientError;
+
+// tslint:disable-next-line:no-any
+export const isExpired = (error: any): error is ExpiredError =>
```
cloudify: why `any` here? can't it be restricted to `RuntimeError`? (also `isTransient`)

also the fact that there's `ErrorTypes.ExpiredError`, `type ExpiredError`, `const ExpiredError`, `const RuntimeError` and `type RuntimeError` looks very confusing and overly complicated

lib/utils/errors.ts
```diff
@@ -50,8 +54,16 @@ export const of = (error: any): RuntimeError =>
         error instanceof Error ? error : undefined
       );
 
-export type RuntimeError = TransientError | PermanentError | UnknowError;
+export type RuntimeError =
+  | TransientError
+  | PermanentError
+  | UnknowError
```
cloudify: typo

lib/created_message_queue_handler.ts
```diff
@@ -373,13 +322,11 @@ export async function index(
       )}`
     );
     // we will never be able to recover from this, so don't trigger a retry
-    return stopProcessing;
+    return;
```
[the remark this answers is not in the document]
gunzip: mm this is an early return (not the last statement)

lib/emailnotifications_queue_handler.ts
```diff
@@ -215,6 +211,18 @@ export async function handleNotification(
 ): Promise<Either<RuntimeError, NotificationEvent>> {
   const { message, notificationId, senderMetadata } = emailNotificationEvent;
 
+  // Check if the message is not expired
+  const errorOrActiveMessage = ActiveMessage.decode(message);
+
+  if (isLeft(errorOrActiveMessage)) {
+    // if the message is expired no more processing is necessary
```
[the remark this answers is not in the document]
gunzip: indeed, so it's not an active message.
at this point we are certain that "message" is a NewMessageWithContent (decoded before from NotificationEvent payload which is passed in input to this method) but it maybe an expired message.

lib/emailnotifications_queue_handler.ts
```diff
@@ -433,75 +391,78 @@ export async function index(
...
-    //
-  } catch (error) {
-    const shouldTriggerARetry = await processRuntimeError(
-      queueService,
-      notificationStatusUpdater,
-      context.bindingData,
-      RuntimeErrorOf(error)
+  )
+    .then(errorOrEmailNotificationEvt =>
+      errorOrEmailNotificationEvt.fold(
```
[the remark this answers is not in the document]
gunzip: then() handles nested promises flattening them. so we return a promise or not, it does not change anything. 
it always return one promise anyway. we may add "await" but that wouldn't change the semantic.

lib/utils/azure_queues.ts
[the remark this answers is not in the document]
gunzip: right, fixed

lib/utils/azure_queues.ts
gunzip: we may do that, but the the caller is responsible to translate an Error in a RunTimeError (calling RuntimeErrorOf)

lib/utils/azure_queues.ts
gunzip: It translates an upcoming Error (or anything that may be throwed inside the code) into a RuntimeError

lib/utils/errors.ts
gunzip: right, I'll try to refactor this using io-ts (we need a hiearchy of types which can be discovered at runtime)
but we'll still got const and type (at least)

gunzip: even using io-ts the situation does not change, see https://gist.github.com/gunzip/ede14c7fd913b913222dd1f109cbda86

what do you suggest to simplify it ?

lib/utils/errors.ts
gunzip: fixed

lib/emailnotifications_queue_handler.ts
```diff
@@ -433,75 +391,78 @@ export async function index(
...
+        async error => {
```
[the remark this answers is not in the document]
gunzip: then flatten and return the nested promise (this should answer both the questions =)

gunzip:
@cloudify 

> disclaimer: I understood 1/3 of this code, it's very complicated :)

indeed, that's the reason for this PR :) 

to make a somewhat complicated error handling easier for those who writes queue message handlers.

the logic is complicated even to describe with words, to summarize:
do something with the message then

1. check is some error occurred
    1. if it's a transient error, try to retry
        1. maybe we got an error while retrying, trying to retry again
      1. try to update the message status (throttled)
          1. maybe we got an error updating the message status, try to retry again
    1. if it's a permanent error or an unknow error, try to update the message status (failed)
        1. maybe we got an error trying to update the message status, try to retry

PR #175, opened 2018-03-29
[#156379054] fix funcpack build with custom webpack config

gunzip:
the actual build fails with 

`Exception while executing function: Functions.AdminApi -> One or more errors occurred. -> TypeError: require is not a function\n`

this fixes an error with funcpack + modules that "gently" hijacks require, see
https://github.com/felixge/node-formidable/issues/337

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,0 +1,10 @@
+/**
+ * Decorates the default webpack configuration (coming from funcpack).
+ * See https://github.com/Azure/azure-functions-pack
+ */
+module.exports = function(config, webpack) {
+  // fix an error with npm modules that call GENTLY.hijack(require)
+  // ie. formidable needed by superagent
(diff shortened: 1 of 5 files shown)


webpack.config.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Decorates the default webpack configuration (coming from funcpack).
+ * See https://github.com/Azure/azure-functions-pack
+ */
+module.exports = function(config, webpack) {
+  // fix an error with npm modules that call GENTLY.hijack(require)
```
cloudify: if there's an open bug/issue somewhere pls link it in the code so in the future we can check if it's been fixed

gunzip: right. note that "casually" in the funcpack readme there's an example with the fix to this bug :)

PR #177, opened 2018-04-04
[#156484880] refactor: avoid taking option as arguments in functions

gunzip:
because option as arguments are code smells

### DIFF
--- a/lib/created_message_queue_handler.ts
+++ b/lib/created_message_queue_handler.ts
@@ -41,13 +41,11 @@ import { NotificationEvent } from "./models/notification_event";
 import { ProfileModel, RetrievedProfile } from "./models/profile";
 
 import { Either, isLeft, left, right } from "fp-ts/lib/Either";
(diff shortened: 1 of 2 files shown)


digitalcitizenship:
## Affected stories

  * ⚙️ [#156484880](https://www.pivotaltracker.com/story/show/156484880): Refactor delle funzioni dell'handler che prendono in input delle Option

<p align="right">
</p>

lib/created_message_queue_handler.ts
```diff
@@ -243,11 +223,10 @@ export async function handleMessage(
     }
   }
 
-  // Try to get email notification metadata from user's profile
-  const maybeNotificationChannelEmail = tryGetEmailNotification(
-    maybeProfile,
-    defaultAddresses
-  );
+  // Try to get email notification metadata
```
cloudify: worth explaining what happens in both cases (email address exists in profile + email not defined)

PR #182, opened 2018-04-06
[#150588220] pushes notifications to a webhook exposed by the API proxy

gunzip:
Since we have two channels now (emai l+ push) the dispatching logic (created_message_queue_handler) changed a little bit. While we used to fail with a permanent error in case no email could be sent. now it just logs a 'warning' for every channel we cannot send the message to then returns an empty object in case no channel can be reached (so queue processing stops).


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  Please include a description of your PR changes.
  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#150588220](https://www.pivotaltracker.com/story/show/150588220): Come SA voglio poter inoltrare gli avvisi all'api proxy, in modo che possa inviare notifiche push a CIT

<p align="right">
</p>

lib/__tests__/created_message_queue_handler.test.ts
```diff
@@ -6,6 +6,7 @@
 // tslint:disable-next-line:no-object-mutation
 process.env = {
   ...process.env,
+  API_PROXY_WEBHOOK_URL: "http://anUrl.com",
```
cloudify: `aUrl`

api/definitions.yaml
```diff
@@ -380,3 +396,8 @@ Timestamp:
   type: string
   format: date-time
   description: A date-time field in ISO-8601 format
+WebUrl:
+  type: string
+  format: uri
+  description: An http(s) url.
+  pattern: "^http(s)?://[-a-zA-Z0-9@:%._+~#=]{2,256}[.][a-z]{2,6}\\\\b([-a-zA-Z0-9@:%_+.~#?&/=]*)$"
```
cloudify: we want this to be always `https`

lib/__tests__/created_message_queue_handler.test.ts
```diff
@@ -6,6 +6,7 @@
 // tslint:disable-next-line:no-object-mutation
 process.env = {
   ...process.env,
+  API_PROXY_WEBHOOK_URL: "http://aUrl.com",
```
cloudify: using `API_PROXY` in this context could be confusing, I would just use something like `WEBHOOK_CHANNEL_URL`

lib/__tests__/created_message_queue_handler.test.ts
```diff
@@ -147,6 +149,8 @@ const anAttachmentMeta = {
   media: "media.json"
 };
 
+const aUrl = "http://aUrl.com" as WebUrl;
```
cloudify: `HttpUrl` would be more correct since we're restricting the scheme

cloudify: or perhaps just URL ?

lib/created_message_queue_handler.ts
```diff
@@ -167,6 +183,49 @@ function getEmailAddressFromProfile(
...
+  const newNotification = createNewNotification(
+    ulidGenerator,
+    newMessageWithContent.fiscalCode,
+    newMessageWithContent.id
+  );
+
+  const errorOrNotification = await lNotificationModel.create(
+    {
+      ...newNotification,
+      channels: {
```
cloudify: we is this attribute plural if it always holds one single channel?

lib/created_message_queue_handler.ts
```diff
@@ -223,56 +287,77 @@ export async function handleMessage(
...
+  //  Return notification events (one for each channel)
+  //
+  const outputBindings: OutputBindings = {
+    emailNotification: maybeEmailNotification.toUndefined(),
+    webhookNotification: maybeWebhookNotification.toUndefined()
   };
 
-  return right({ emailNotification });
+  // avoid to enqueue messages for non existing notifications
+  return right(withoutUndefinedValues(outputBindings));
```
cloudify: instead of transforming the `Option`s to `.toUndefined()` and then calling `withoutUndefinedValues` you could skip the `.toUndefined()` and filter with `isSome`

lib/models/notification.ts
```diff
@@ -69,22 +70,39 @@ export const EmailNotification = t.interface({
...
+export const NotificationChannelWebhook = t.interface({
+  url: WebUrl
+});
+export type NotificationChannelWebhook = t.TypeOf<
+  typeof NotificationChannelWebhook
+>;
+
+export const WebhookNotification = t.interface({
+  ...NotificationBase.props,
+  channels: t.interface({
```
cloudify: I don't understand why we need `channels: {}` with a single key in it

lib/__tests__/created_message_queue_handler.test.ts
gunzip: cannot be "URL" as it conflicts with nodejs URL type

cloudify: can't we use that instead of creating a new type?

gunzip: probably we can. I guess that to that we must add the type to the api definition template (nunjucks) as we cannot directly reference a "native" type in the api specs. maybe with another pr ?

lib/created_message_queue_handler.ts
gunzip: not really clear, may you provide an example ?

cloudify: sorry I always forget there's not filter on maps in ts, never mind - what I meant was basically having a method like `withoutUndefinedValues` that filtered out empty `Option`s

api/definitions.yaml
gunzip: I've added a custom type to the italia-ts-common repo:

https://github.com/teamdigitale/italia-ts-commons/pull/10

lib/__tests__/created_message_queue_handler.test.ts
gunzip: done

lib/created_message_queue_handler.ts
gunzip: renamed

lib/models/notification.ts
gunzip: renamed

lib/models/notification.ts
```diff
@@ -49,41 +49,41 @@ export const NotificationBase = t.interface({
...
-  t.interface({
-    addressSource: NotificationAddressSource,
-    toAddress: EmailAddress
-  }),
-  t.partial({
-    fromAddress: EmailAddress
-  })
-]);
+export const NotificationChannelEmail = t.interface({
+  [NotificationChannelEnum.EMAIL]: t.intersection([
```
cloudify: I don't understand, this type describes a `NotificationChannelEmail`, why it's an object with a single constant key? (`NotificationChannelEnum.EMAIL`) - why `NotificationChannelEmail` is not `t.intersection([...])` ? (same below for `NotificationChannelWebhook`)

lib/models/notification.ts
```diff
@@ -97,15 +97,7 @@ export type NotificationChannelMeta = t.TypeOf<typeof NotificationChannelMeta>;
...
-export const Notification = t.intersection([
-  NotificationBase,
-  t.interface({
-    channel: t.partial({
-      [NotificationChannelEnum.EMAIL]: NotificationChannelEmail,
-      [NotificationChannelEnum.WEBHOOK]: NotificationChannelWebhook
-    })
-  })
-]);
+export const Notification = t.union([WebhookNotification, EmailNotification]);
```
cloudify: I think it made much more sense before than now

gunzip: needs some context : ) 

most of the oddities here come from the fact that we started with a "fat" notification model that contained all metadata for every channel.  now we handle only specific kinds of notifications
(ie. email *or* webhook).

so regarding "why this property is called *channels* and not *channel* ?": because renaming it, despite being more correct, would be a breaking change (a change not exposed to the external API, but will break the way we retrieve notifications object from CosmosDB).

in case we rename it,it's worth to refactor the whole notification model to adhere much more to the new requirements. for example the old way (which you state that "made much more sense") is plain wrong:

```ts
export const Notification =t.interface({
 NotificationBase
  t.interface({
    channel: t.partial({	
      [NotificationChannelEnum.EMAIL]: NotificationChannelEmail	
      [NotificationChannelEnum.WEBHOO]: NotificationChannelWebhook
    })	
  })	
 ]);
```

We cannot have both `NotificationChannelEmail` *and* `NotificationChannelWebhook` (this was possible with the old fat model), so `Notification` is infact a union type. 

I've restructured the types again a little bit to "flatten" the metadata avoiding the 
"single constant key" and using it to discriminate the tagged union (= Notification).

lib/models/notification.ts
```diff
@@ -49,55 +49,48 @@ export const NotificationBase = t.interface({
...
+  url: HttpsUrl
 });
 export type NotificationChannelWebhook = t.TypeOf<
   typeof NotificationChannelWebhook
 >;
 
 export const WebhookNotification = t.interface({
   ...NotificationBase.props,
-  channel: NotificationChannelWebhook
+  channel: NotificationChannelWebhook,
```
cloudify: same here, it looks like having the channel specific attributes inside `channel` can be avoided, getting rid of the extra type `NotificationChannelWebhook`

cloudify: same here, it looks like having the channel specific attributes inside `channel` can be avoided, getting rid of the extra type `NotificationChannelWebhook`

gunzip: two reasons. 1. there are some functions (like [this one](https://github.com/teamdigitale/digital-citizenship-functions/blob/97279ea7dd5ce53396ad30afb179de56e72d9c1f/lib/created_message_queue_handler.ts#L161)) that works on the channel metadata. we would need to define a type for these functions which has the same shape of NotificationChannelWebhook 2. [toBaseType](https://github.com/teamdigitale/digital-citizenship-functions/pull/182/files#diff-f77a48800133699f87c40a0d01fef39eR138) is supposed to return a "complete" notification (so it extracts the channel metadata aside NotificationBase properties). if we flatten the metadata we will complicate the logic to extract a Notification from a RetrievedNotification.

PR #209, opened 2018-05-04
[#157308219] fix: superagent mock

gunzip:
Actual implmentation is clumsy and not safe to run between tests (mutates global state)

Replaced with this two line implementation: 
https://github.com/teamdigitale/digital-citizenship-functions/pull/209/files#diff-745f8e09108272a03a8fb116f2f16e7bR16

which is also more flexible.


digitalcitizenship:
## Affected stories

  * 🐞 [#157308219](https://www.pivotaltracker.com/story/show/157308219): Fix superagent mock

<p align="right">
</p>

lib/utils/__tests__/mailup.test.ts
```diff
@@ -4,37 +4,20 @@ import Mail = require("nodemailer/lib/mailer");
...
-  (superagent as any).__resetMocks();
-};
+afterEach(() => {
+  jest.restoreAllMocks();
+  jest.resetAllMocks();
+});
+
+const mockSuperagentResponse = (response: any) =>
+  // tslint:disable-next-line:no-object-mutation
+  (Object.getPrototypeOf(request("", "")).send = jest.fn()).mockReturnValueOnce(
```
cloudify: I can't even understand what this does, please explain it very well :)

gunzip: added docs

lib/utils/__tests__/mailup.test.ts
```diff
@@ -13,9 +13,13 @@ afterEach(() => {
...
 });
 
+// as superagent does not export request methods directly
+// we must override the superagent.Request prototype
+// so we can set up our jest mock to use it instead
+// of the send() method
 const mockSuperagentResponse = (response: any) =>
   // tslint:disable-next-line:no-object-mutation
-  (Object.getPrototypeOf(request("", "")).send = jest.fn()).mockReturnValueOnce(
+  ((superagent as any).Request.prototype.send = jest.fn()).mockReturnValueOnce(
```
cloudify: thx, I would just split this line for readability:

```ts
const sendMock = jest.fn();
(superagent as any).Request.prototype.send = sendMock;
sendMock.mockReturnValueOnce(...)
```

gunzip: another option is to just use
```ts
const getSendMock = () => (superagent as any).Request.prototype.send
```

then use it inside tests with 
```
getSendMock().mockReturnValue(...)
```

cloudify: `getSendMock` can be misleading as it creates a new mock each time, perhaps `newSendMock` or `initSendMock`

anyway, just make the code cleaner by splitting the creation of the mock from its usage

gunzip: done

PR #220, opened 2018-05-10
[#157494849] fix notification model

gunzip:
Introducing webhook notifications caused a regression. We used to create one notification object for each message (1-to-1) while now we create a notification object for each channel. This PR reverts the notification model changes

### DIFF
--- a/lib/__tests__/created_message_queue_handler.test.ts
+++ b/lib/__tests__/created_message_queue_handler.test.ts
@@ -27,7 +27,8 @@ import {
   EmailNotification,
   NewNotification,
   NotificationAddressSourceEnum,
-  NotificationModel
+  NotificationModel,
+  WebhookNotification
 } from "../models/notification";
 import { ProfileModel, RetrievedProfile } from "../models/profile";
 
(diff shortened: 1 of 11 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#157494849](https://www.pivotaltracker.com/story/show/157494849): Il modello notification deve esser 1-a-1 con il messaggio

<p align="right">
</p>

lib/__tests__/created_message_queue_handler.test.ts
```diff
@@ -62,16 +63,28 @@ afterEach(() => {
...
 const aWrongFiscalCode = "FRLFRC74E04B157" as FiscalCode;
 const anEmail = "x@example.com" as EmailString;
+const aUrl = "http://aUrl.com" as HttpsUrl;
 
 const anEmailNotification: EmailNotification = {
-  channel: {
-    addressSource: NotificationAddressSourceEnum.PROFILE_ADDRESS,
-    // fromAddress: anEmail,
-    toAddress: anEmail
+  channels: {
```
gunzip: that's why it was called "channels" (plural): because we have one notification object with *all* channels metadata for each message : )

lib/__tests__/created_message_queue_handler.test.ts
```diff
@@ -419,6 +432,59 @@ describe("handleMessage", () => {
     }
   );
 
+  it(
+    "should create a webhook notification if a profile exists for " +
```
gunzip: aside, added a test for webhook notifications

PR #226, opened 2018-05-14
[#157387718] WIP: service blacklist

gunzip:
## Abstract

- adds a new property "blocked_channels" to the profile model 
`{ serviceId: [ array of blocked channels ]}`

- adds logic to skip blocked channel into created_message_queue_handler

profile model: 

```
{
    "blocked_channels": {
        "5a25abf4fcc89605c082f042c49a": [
            "EMAIL",
            "WEBHOOK"
        ]
    },
    "email": "xxxxxxxxx@xxxxxxx.com",
    "is_inbox_enabled": true,
    "is_webhook_enabled": true,
    "preferred_languages": [
        "it_IT"
    ],
    "version": 46
}

```

There at least are a couple of possible implementation. The one in "Scenario 1" is implemented in the current PR.

### Scenario 1

We block notifications in the `created_message_queue_handler`, before dispatching the messages to the channel queue handlers.

**Pros**: efficient as it avoids to send notifications to the channels queue handlers
**Cons**: there are some chances that the "blocked" messages are dispatched to the user in case the act of block a channel occurs *after* the message is already dispatched to the specific channel queue handler: these handlers will send the messages without any further check.

### Scenario 2

We dispatch the notifications to the specific channel queue handlers (anyway), only then we check if the user has blocked the specific channel.

**Pros**: no chance the user wil receive any notification after blocking any channel
**Cons**: require an extra call to retrieve the user's profile in the specific channel queue handler, pollute channel queues with notifications that won't be sent

## Decisions to take

What to return in the notification-status for blocked messages ? (actually we do not save any status, which is the same behavior we get when a notification is not sent for any other reason).

## TODO

- add tests

### DIFF
--- a/lib/created_message_queue_handler.ts
+++ b/lib/created_message_queue_handler.ts
@@ -226,6 +226,7 @@ async function createNotification(
  *          a TransientError in case of recoverable errors
  *          a PermanentError in case of unrecoverable error
  */
+// tslint:disable-next-line:cognitive-complexity
 export async function handleMessage(
   lProfileModel: ProfileModel,
   lMessageModel: MessageModel,
@@ -313,7 +314,7 @@ export async function handleMessage(
 
   if (isEmailBlockedForService) {
     winston.debug(
-      `User has blocked email notifications for this serviceId|${
+      `handleMessage|User has blocked email notifications for this serviceId|${
         newMessageWithContent.fiscalCode
       }:${newMessageWithContent.senderServiceId}`
     );
@@ -333,7 +334,7 @@ export async function handleMessage(
 
   if (noEmailAddressFound) {
     winston.debug(
-      `Fiscal code has no associated email address and no default email address was provided|${
+      `handleMessage|User profile has no email address set and no default address was provided|${
         newMessageWithContent.fiscalCode
       }`
     );
@@ -350,7 +351,7 @@ export async function handleMessage(
 
   if (isWebhookBlockedForService) {
     winston.debug(
-      `User has blocked webhook notifications for this serviceId|${
+      `handleMessage|User has blocked webhook notifications for this serviceId|${
         newMessageWithContent.fiscalCode
       }:${newMessageWithContent.senderServiceId}`
     );
@@ -377,7 +378,7 @@ export async function handleMessage(
       PermanentError(
         `No channels configured for the user ${
           newMessageWithContent.fiscalCode
-        }`
+        } and no default address provided`
       )
     );
   }
--- a/lib/created_message_queue_handler.ts
+++ b/lib/created_message_queue_handler.ts
@@ -17,7 +17,16 @@ import { configureAzureContextTransport } from "./utils/logging";
 
 import * as documentDbUtils from "./utils/documentdb";
 
-import { fromNullable, isNone, none, Option, some } from "fp-ts/lib/Option";
+import { Set } from "json-set-map";
+
+import {
+  fromNullable,
+  fromPredicate,
+  isNone,
+  none,
+  Option,
+  some
+} from "fp-ts/lib/Option";
 
 import {
   BlobService,
@@ -51,6 +60,7 @@ import { MessageStatusValueEnum } from "./api/definitions/MessageStatusValue";
 import { NotificationChannelEnum } from "./api/definitions/NotificationChannel";
 
 import { withoutUndefinedValues } from "italia-ts-commons/lib/types";
+import { BlockedChannelEnum } from "./api/definitions/BlockedChannel";
 import { HttpsUrl } from "./api/definitions/HttpsUrl";
 import { CreatedMessageEventSenderMetadata } from "./models/created_message_sender_metadata";
 import {
@@ -247,14 +257,36 @@ export async function handleMessage(
   // query succeeded, we may have a profile
   const maybeProfile = errorOrMaybeProfile.value;
 
-  // whether the recipient wants us to store the message content
-  const isMessageStorageEnabled = maybeProfile.exists(
-    profile => profile.isInboxEnabled === true
+  // channels ther user has blocked for this sender service
+  const blockedChannels = new Set(
+    maybeProfile
+      .chain(profile =>
+        fromNullable(profile.blockedChannels).map(
+          bc => bc[newMessageWithContent.senderServiceId]
+        )
+      )
+      .getOrElse([])
+  );
+
+  winston.debug(
+    "handleMessage|Blocked Channels(%s): %s",
+    newMessageWithContent.fiscalCode,
+    JSON.stringify(blockedChannels)
   );
 
   //
   //  Inbox storage
   //
+
+  // check if the user has blocked webhook notifications sent from this service
+  const userHasBlockedMessageStorage = blockedChannels.has(
+    BlockedChannelEnum.INBOX
+  );
+
+  // whether the recipient wants us to store the message content
+  const isMessageStorageEnabled =
+    !userHasBlockedMessageStorage &&
+    maybeProfile.exists(profile => profile.isInboxEnabled === true);
 
   if (isMessageStorageEnabled) {
     // If the recipient wants to store the messages
@@ -276,6 +308,21 @@ export async function handleMessage(
   //  Email notification
   //
 
+  // check if the user has blocked emails sent from this service
+  // 'some(true)' in case we must send the notification by email
+  // 'none' in case the user has blocked the email channel
+  const maybeIsUnblockedEmail = fromPredicate(
+    (hasBlockedEmail: boolean) => !hasBlockedEmail
+  )(blockedChannels.has(BlockedChannelEnum.EMAIL));
+
+  if (isNone(maybeIsUnblockedEmail)) {
+    winston.warn(
+      `User has blocked email notifications for this serviceId|${
+        newMessageWithContent.fiscalCode
+      }:${newMessageWithContent.senderServiceId}`
+    );
+  }
+
   const maybeEmailNotification = maybeProfile
     // try to get the destination email address from the user's profile
     .chain(getEmailAddressFromProfile)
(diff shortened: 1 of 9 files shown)


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#157387718](https://www.pivotaltracker.com/story/show/157387718): Come CIT voglio poter decidere di non ricevere notifiche da uno specifico servizio di un SAS

<p align="right">
</p>

cloudify:
I think Scenario 1 is a reasonable approach.

About what we return, I'd say we keep it as is. The sender shouldn't probably know anything about what happens after he sends the message.

lib/created_message_queue_handler.ts
```diff
@@ -247,15 +261,36 @@ export async function handleMessage(
...
+    JSON.stringify(blockedChannels)
   );
 
   //
   //  Inbox storage
   //
 
+  // check if the user has blocked inbox message storage from this service
+  const userHasBlockedMessageStorage = blockedChannels.has(
+    BlockedChannelEnum.INBOX
```
cloudify: how is this option exposed to the user?

lib/created_message_queue_handler.ts
```diff
@@ -247,15 +261,36 @@ export async function handleMessage(
...
+    newMessageWithContent.fiscalCode,
+    JSON.stringify(blockedChannels)
   );
 
   //
   //  Inbox storage
   //
 
+  // check if the user has blocked inbox message storage from this service
+  const userHasBlockedMessageStorage = blockedChannels.has(
```
cloudify: just to be clear (while reading the code I forgot this is for a specific service) `userHasBlockedMessageStorageForService`

lib/created_message_queue_handler.ts
cloudify: we need to clarify this from a UX point of view: does it makes sense to disable WEBHOOK and not INBOX? shouldn't they go together?

lib/created_message_queue_handler.ts
```diff
@@ -276,6 +311,21 @@ export async function handleMessage(
   //  Email notification
   //
 
+  // check if the user has blocked emails sent from this service
+  // 'some(true)' in case we must send the notification by email
+  // 'none' in case the user has blocked the email channel
+  const maybeIsUnblockedEmail = fromPredicate(
```
cloudify: negative booleans are hard to keep track of (mentally): `maybeIsUnblockedEmail` -> `maybeIsEmailAllowed`

lib/created_message_queue_handler.ts
```diff
@@ -276,6 +311,21 @@ export async function handleMessage(
...
+    (hasBlockedEmail: boolean) => !hasBlockedEmail
+  )(blockedChannels.has(BlockedChannelEnum.EMAIL));
+
+  if (isNone(maybeIsUnblockedEmail)) {
```
cloudify: try to avoid logic that is there only for logging purposes, perhaps you can move the logging to a code branch that gets executed only when `maybeIsUnblockedEmail` is empty

lib/created_message_queue_handler.ts
```diff
@@ -291,22 +341,40 @@ export async function handleMessage(
     );
   }
 
+  const maybeUnblockedEmailNotification = maybeIsUnblockedEmail.chain(
```
cloudify: +comment since it's hard to follow what's going on

lib/created_message_queue_handler.ts
```diff
@@ -291,22 +341,40 @@ export async function handleMessage(
...
+    _ => maybeEmailNotification
+  );
+
   //
   //  Webhook notification
   //
 
+  // check if the user has blocked webhook notifications sent from this service
+  const isBlockedWebhook = blockedChannels.has(BlockedChannelEnum.WEBHOOK);
```
cloudify: think of boolean identifiers as questions: is webhook blocked? -> `isWebhookBlocked` (also for consistency with `isWebhookEnabled`)

lib/created_message_queue_handler.ts
```diff
@@ -276,6 +311,21 @@ export async function handleMessage(
...
 
+  // check if the user has blocked emails sent from this service
+  // 'some(true)' in case we must send the notification by email
+  // 'none' in case the user has blocked the email channel
+  const maybeIsUnblockedEmail = fromPredicate(
+    (hasBlockedEmail: boolean) => !hasBlockedEmail
+  )(blockedChannels.has(BlockedChannelEnum.EMAIL));
+
+  if (isNone(maybeIsUnblockedEmail)) {
+    winston.warn(
```
cloudify: is it something we need to be concerned about? this is probably `debug` level

lib/created_message_queue_handler.ts
```diff
@@ -291,22 +341,40 @@ export async function handleMessage(
...
+  );
+
   //
   //  Webhook notification
   //
 
+  // check if the user has blocked webhook notifications sent from this service
+  const isBlockedWebhook = blockedChannels.has(BlockedChannelEnum.WEBHOOK);
+
+  if (isBlockedWebhook) {
```
cloudify: similar comment as above

lib/created_message_queue_handler.ts
```diff
@@ -291,22 +341,40 @@ export async function handleMessage(
...
   // whether the recipient wants us to send notifications to the app backend
-  const isWebhookEnabled = maybeProfile.exists(
-    profile => profile.isWebhookEnabled === true
-  );
+  const isWebhookEnabled =
+    !isBlockedWebhook &&
+    maybeProfile.exists(profile => profile.isWebhookEnabled === true);
 
-  const maybeWebhookNotification = isWebhookEnabled
+  const maybeUnblockedWebhookNotification = isWebhookEnabled
```
cloudify: Unblocked/NotBlocked -> Allowed (prefer "positive" identifiers)

lib/models/profile.ts
```diff
@@ -15,13 +15,26 @@ import { Option } from "fp-ts/lib/Option";
...
 
+const ProfileBlockedChannels = t.dictionary(
+  ServiceId,
+  readonlySetType(BlockedChannel, "ProfileBlockedChannel"),
+  "ProfileBlockedChannels"
+);
+
+// typescript does not allow to have ServiceId as key type here
+export interface IProfileBlockedChannels {
+  readonly [serviceId: string]: ReadonlySet<BlockedChannel>;
```
cloudify: why not `ServiceId` instead of `string`?

lib/created_message_queue_handler.ts
gunzip: another option is to replce "INBOX" with "ALL" and block all notification channels *and* the inbox.
in other words, if the inbox (message storage) is blocked, then every channel is blocked. 

cc @matteodesanti

lib/models/profile.ts
gunzip: because typescript does not allow to have ServiceId as key type (only strings, as stated by the comment above : )

lib/created_message_queue_handler.ts
cloudify: yes, conceptually the inbox is not a notification channel (the inbox is orthogonal to the notifications) - what do you think about something like:

```ts
type BlockedInboxOrChannels = { [k: Channel]: true } | "INBOX";
```
`INBOX` implies inbox and all channels are blocked

gunzip: nice but note that openapi 2.0 spec (the version we use) does not support "oneOf" in the schema and so doesn't our typescript generator (italia-utils). it means that

1. we cannot have that type generated from the swagger specs
2. actually there's no way to enforce this in clients

cloudify: but that's an opportunity to add a new feature to our tool! 😸

gunzip: as a workaround we can just settle to union("inbox", "webhook", "email") and handle the logic server side (more or less like the actual logic but implying "webhook", "email" in case inbox is set)

cloudify: yes, less elegant (the type doesn't fully describe the logic) but I guess that's the only choice

lib/created_message_queue_handler.ts
gunzip: it's used afterwards

gunzip:
> About what we return, I'd say we keep it as is. The sender shouldn't probably know anything about what happens after he sends the message.

That's right but we may need to answer the question "why that notification we sent in the past wasn't delivered ?"

As the user may unblock services, the information that the notification wasn't delivered due to the user's past blacklist gets lost. That's why, I guess, we should save the NotificationStatus (ie. "blocked") anyway.

@cloudify

cloudify:
> That's right but we may need to answer the question "why that notification we sent in the past wasn't delivered ?"

why we need to answer that question? notification may not be delivered for a multitude of reasons out of our control (e.g. email flagged as spam)

lib/created_message_queue_handler.ts
```diff
@@ -17,6 +17,8 @@ import { configureAzureContextTransport } from "./utils/logging";
 
 import * as documentDbUtils from "./utils/documentdb";
 
+import { Set } from "json-set-map";
```
cloudify: We have [this io-ts set](https://teamdigitale.github.io/italia-ts-commons/modules/_types_.html#readonlysettype)

lib/created_message_queue_handler.ts
```diff
@@ -219,6 +226,7 @@ async function createNotification(
  *          a TransientError in case of recoverable errors
  *          a PermanentError in case of unrecoverable error
  */
+// tslint:disable-next-line:cognitive-complexity
```
cloudify: 🆘

lib/created_message_queue_handler.ts
```diff
@@ -295,23 +344,41 @@ export async function handleMessage(
...
+    );
+  }
+
   // whether the recipient wants us to send notifications to the app backend
-  const isWebhookEnabled = maybeProfile.exists(
-    profile => profile.isWebhookEnabled === true
-  );
+  const isWebhookEnabled =
+    !isWebhookBlockedForService &&
+    maybeProfile.exists(profile => profile.isWebhookEnabled === true);
```
cloudify: just for added readability, I would assign this value first to a const, e.g.:

```ts
const isWebhookEnabledInProfile = maybeProfile.exists(profile => profile.isWebhookEnabled === true);
const isWebhookEnabled = isWebhookEnabledInProfile && !isWebhookBlockedForService;
```

lib/created_message_queue_handler.ts
gunzip: yes, but that's  not what we use in other models (ie. service with authorized_cidrs). we may consider a broader refactor using io-ts sets instead of this one (but we must check that io-ts sets are serializable first)

lib/created_message_queue_handler.ts
gunzip: fixed. note that this come from the fact that every block ```if(...) { winston.debug }``` increases the cognitive-complexity : )

PR #232, opened 2018-05-21
[#157552795] save service list and expose endpoint

gunzip:
embeds this: https://github.com/teamdigitale/digital-citizenship-functions/pull/250

### DIFF
--- a/api/public_api_v1.yaml
+++ b/api/public_api_v1.yaml
@@ -35,6 +35,37 @@ paths:
         '404':
           description: No service found for the provided ID.
       parameters: []
+  "/services":
+    get:
+      operationId: getServices
+      summary: Get Services by recipient
+      description:  |-
+        Returns the service IDs of all the services that have contacted the recipient,
+        identified by the provided fiscal code, at least once.
+      responses:
+        '200':
+          description: Found.
+          schema:
+            allOf:
+            - type: object
+              properties:
+                items:
+                  type: array
+                  items:
+                    serviceId:
+                      $ref: "#/definitions/ServiceId"
+            - $ref: "#/definitions/PaginationResponse"
+          examples:
+            application/json:
+              page_size: 1
+              items:
+                - service_id: 2b3e728c1a5d1efa035c
+      parameters:
+      - name: recipient
+        in: query
+        $ref: "#/definitions/FiscalCode"
+        required: true
+        description: The ID of an existing Service.
   "/messages/{fiscal_code}/{id}":
     parameters:
     - $ref: "#/parameters/FiscalCode"
--- a/lib/controllers/services.ts
+++ b/lib/controllers/services.ts
@@ -152,7 +152,7 @@ export function GetService(serviceModel: ServiceModel): express.RequestHandler {
  * Returns the serviceId for all the Services that have sent
  * at least one notification to the recipient with the provided fiscalCode.
  */
-export function GetSenderServicesHandler(
+export function GetServicesByRecipientHandler(
   senderServiceModel: SenderServiceModel
 ): IGetSenderServicesHandler {
   return async (_, __, ___, fiscalCode) => {
@@ -172,16 +172,16 @@ export function GetSenderServicesHandler(
 /**
  * Wraps a GetSenderServices handler inside an Express request handler.
  */
-export function GetSenderServices(
+export function GetServicesByRecipient(
   serviceModel: ServiceModel,
   senderServiceModel: SenderServiceModel
 ): express.RequestHandler {
-  const handler = GetSenderServicesHandler(senderServiceModel);
+  const handler = GetServicesByRecipientHandler(senderServiceModel);
   const azureUserAttributesMiddleware = AzureUserAttributesMiddleware(
     serviceModel
   );
   const middlewaresWrap = withRequestMiddlewares(
-    AzureApiAuthMiddleware(new Set([UserGroup.ApiFullProfileRead])),
+    AzureApiAuthMiddleware(new Set([UserGroup.ApiServiceByRecipientQuery])),
     ClientIpMiddleware,
     azureUserAttributesMiddleware,
     RequiredQueryParamMiddleware("recipient", FiscalCode)
--- a/lib/controllers/services.ts
+++ b/lib/controllers/services.ts
@@ -46,7 +46,6 @@ import {
   IAzureUserAttributes
 } from "../utils/middlewares/azure_user_attributes";
 
-import { pick } from "italia-ts-commons/lib/types";
 import { FiscalCode } from "../api/definitions/FiscalCode";
 import { ServiceId } from "../api/definitions/ServiceId";
 import { SenderServiceModel } from "../models/sender_service";
@@ -72,7 +71,7 @@ type IGetServiceHandler = (
 
 type IGetSenderServicesHandlerRet =
   | IResponseSuccessJsonIterator<{
-      readonly serviceId: ServiceId;
+      readonly service_id: ServiceId;
     }>
   | IResponseErrorQuery;
 
@@ -162,7 +161,9 @@ export function GetSenderServicesHandler(
     );
     const senderServicesIterator = mapResultIterator(
       retrievedServicesIterator,
-      service => pick(["serviceId"], service)
+      service => ({
+        service_id: service.serviceId
+      })
     );
     return ResponseSuccessJsonIterator(senderServicesIterator);
   };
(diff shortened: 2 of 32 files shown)


lib/controllers/services.ts
```diff
@@ -56,9 +64,23 @@ type IGetServiceHandler = (
...
   clientIp: ClientIp,
   userAttributes: IAzureUserAttributes,
-  serviceId: NonEmptyString
+  serviceId: ServiceId
 ) => Promise<IGetServiceHandlerRet>;
 
+type IGetSenderServicesHandlerRet =
+  | IResponseSuccessJsonIterator<{
+      readonly serviceId: ServiceId;
+      readonly fiscalCode: FiscalCode;
```
cloudify: is this the service fiscal code? (in that case, why it's needed by the client?)

gunzip: it is not, I'll remove it

lib/created_message_queue_handler.ts
```diff
@@ -238,7 +253,7 @@ export async function handleMessage(
   const defaultAddresses = fromNullable(createdMessageEvent.defaultAddresses);
   const senderMetadata = createdMessageEvent.senderMetadata;
 
-  // async fetch of profile data associated to the fiscal code the message
+  // fetch user's profile associated to the fiscal code the message
```
cloudify: ...to the fiscal code of the recipient of the message?

lib/created_message_queue_handler.ts
```diff
@@ -285,8 +299,7 @@ export async function handleMessage(
     maybeProfile.exists(profile => profile.isInboxEnabled === true);
 
   if (isMessageStorageEnabledAndAllowedForService) {
-    // If the recipient wants to store the messages
-    // we add the content of the message to the blob storage for later retrieval.
+    // Save message content to the blob storage.
```
cloudify: Save the message content...

lib/models/sender_service.ts
```diff
@@ -0,0 +1,135 @@
...
+import { DocumentDbModel } from "../utils/documentdb_model";
+
+import { FiscalCode } from "../api/definitions/FiscalCode";
+
+import { DateFromString } from "italia-ts-commons/lib/dates";
+import { NonEmptyString } from "italia-ts-commons/lib/strings";
+import { ServiceId } from "../api/definitions/ServiceId";
+
+export const SENDER_SERVICE_COLLECTION_NAME = "sender-services";
+export const SENDER_SERVICE_MODEL_ID_FIELD = "fiscalCode";
```
cloudify: what's the difference between `ID` and `PK`?

lib/models/sender_service.ts
```diff
@@ -0,0 +1,135 @@
...
+export const SENDER_SERVICE_MODEL_PK_FIELD = "fiscalCode";
+
+export const SenderService = t.interface({
+  createdAt: DateFromString,
+  fiscalCode: FiscalCode,
```
cloudify: `fiscalCode` is not needed here (?)

cloudify: ok sorry this is the fiscal code of the recipient... we should probably give to these attributes a more specific name, like `recipient: FiscalCode` and `sender: ServiceId`

lib/public_api_v1.ts
```diff
@@ -124,6 +139,11 @@ app.post("/api/v1/debug", debugHandler);
 
 app.get("/api/v1/services/:serviceid", GetService(serviceModel));
 
+app.get(
+  "/api/v1/sender-services/:fiscalCode",
```
cloudify: this pattern doesn't work here, this means that `fiscalCode` is an identifier for the `sender-services` resource - in REST terms `fiscalCode` is a query pattern for retrieving `services` (the resource we're returning)

perhaps something like `/profile/:fiscalCode/senders` could work better or, if we want to avoid nested resources: `/services/for-recipient/:fiscalCode`

lib/models/sender_service.ts
gunzip: id is the name of the id field (ie. "id") which is used by CosmosDB as the row index.
partitionKey is needed during CosmosDB queries to let it find the partition.

lib/utils/documentdb.ts
```diff
@@ -449,6 +449,45 @@ export async function iteratorToArray<T>(
...
+  partitionKey: string
+): Promise<Either<DocumentDb.QueryError, T & DocumentDb.RetrievedDocument>> {
+  return new Promise(resolve => {
+    client.upsertDocument(
+      collectionUri.uri,
+      document,
+      {
+        partitionKey
+      },
+      /* tslint:disable-next-line:no-identical-functions */
```
cloudify: weird that it's complaining about this

lib/utils/documentdb.ts
```diff
@@ -449,6 +449,45 @@ export async function iteratorToArray<T>(
   return iterate([]);
 }
 
+/**
+ * Replaces an existing document with a new one
```
cloudify: this should probably go in a separated PR, with its own tests

lib/models/sender_service.ts
gunzip: if the path would be /services?recipient=fiscalCode I think that we should maintain serviceId (as a Service sub-property)

cloudify: ok for me for serviceId in that context is not ambiguous

digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  Please include a description of your PR changes.
  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#157552795](https://www.pivotaltracker.com/story/show/157552795): Come CIT voglio poter visualizzare la lista dei servizi che mi hanno contattato, in modo da poter scegliere quelli di cui voglio bloccare le notifiche

<p align="right">
</p>

lib/controllers/services.ts
```diff
@@ -123,3 +148,46 @@ export function GetService(serviceModel: ServiceModel): express.RequestHandler {
...
+export function GetSenderServices(
+  serviceModel: ServiceModel,
+  senderServiceModel: SenderServiceModel
+): express.RequestHandler {
+  const handler = GetSenderServicesHandler(senderServiceModel);
+  const azureUserAttributesMiddleware = AzureUserAttributesMiddleware(
+    serviceModel
+  );
+  const middlewaresWrap = withRequestMiddlewares(
+    AzureApiAuthMiddleware(new Set([UserGroup.ApiFullProfileRead])),
```
cloudify: we need a new API permission for this operation, something like `ApiServiceByRecipientQuery`

api/public_api_v1.yaml
```diff
@@ -35,6 +35,37 @@ paths:
         '404':
           description: No service found for the provided ID.
       parameters: []
+  "/services":
+    get:
+      operationId: getServices
```
cloudify: this operation is more specific, like `getServicesByRecipient`

lib/controllers/services.ts
```diff
@@ -123,3 +148,46 @@ export function GetService(serviceModel: ServiceModel): express.RequestHandler {
     )
   );
 }
+
+/**
+ * Returns the serviceId for all the Services that have sent
+ * at least one notification to the recipient with the provided fiscalCode.
+ */
+export function GetSenderServicesHandler(
```
cloudify: I would align the method name to the operationId of the API spec (`getServicesByRecipient`)

api/public_api_v1.yaml
gunzip: according to the swagger 2.0 specs there's no way to map multiple operations on the same endpoint which differs only by a query string parameter. if you want to name this getServicesByRecipient then a /services-by-recipients endpoint would be a better fit

cloudify: I prefer to keep the resource as is, thanks for checking

lib/controllers/services.ts
```diff
@@ -172,16 +172,16 @@ export function GetSenderServicesHandler(
...
   senderServiceModel: SenderServiceModel
 ): express.RequestHandler {
-  const handler = GetSenderServicesHandler(senderServiceModel);
+  const handler = GetServicesByRecipientHandler(senderServiceModel);
   const azureUserAttributesMiddleware = AzureUserAttributesMiddleware(
     serviceModel
   );
   const middlewaresWrap = withRequestMiddlewares(
-    AzureApiAuthMiddleware(new Set([UserGroup.ApiFullProfileRead])),
+    AzureApiAuthMiddleware(new Set([UserGroup.ApiServiceByRecipientQuery])),
```
cloudify: make sure this is added to the API key of the app backend

PR #233, opened 2018-05-23
[#157800792] feat: add smtp nodemailer transport

gunzip:
bumped italia-ts-commons to bring UrlFromString


gunzip:
not needed atm

PR #237, opened 2018-05-27
[#151202533] add telemetry and insights

gunzip:
Added tracking of:

for each queue handler:
- number of TransientErrors (retry) + time elapsed from message creation
- number of PermanentErrors + time elapsed from message creation
- number of Sucessful processing + time elapsed from message creation

- duration of calls to email provider (and failures)
- duration of calls to webhook endpoint (and failures)

that would be useful to monitor queues lenght as well, but that would require a dedicated function.

### DIFF
--- a/lib/controllers/messages.ts
+++ b/lib/controllers/messages.ts
@@ -108,7 +108,7 @@ import { TimeToLiveSeconds } from "../api/definitions/TimeToLiveSeconds";
 import { MessageStatusModel } from "../models/message_status";
 import { NotificationStatusModel } from "../models/notification_status";
 import {
-  CustomTelemetryClient,
+  CustomTelemetryClientFactory,
   diffInMilliseconds
 } from "../utils/application_insights";
 
@@ -324,7 +324,7 @@ async function getMessageNotificationStatuses(
  * Returns a type safe CreateMessage handler.
  */
 export function CreateMessageHandler(
-  getCustomTelemetryClient: CustomTelemetryClient,
+  getCustomTelemetryClient: CustomTelemetryClientFactory,
   messageModel: MessageModel,
   generateObjectId: ObjectIdGenerator
 ): ICreateMessageHandler {
@@ -510,7 +510,7 @@ export function CreateMessageHandler(
  * Wraps a CreateMessage handler inside an Express request handler.
  */
 export function CreateMessage(
-  getCustomTelemetryClient: CustomTelemetryClient,
+  getCustomTelemetryClient: CustomTelemetryClientFactory,
   serviceModel: ServiceModel,
   messageModel: MessageModel
 ): express.RequestHandler {
--- a/lib/controllers/messages.ts
+++ b/lib/controllers/messages.ts
@@ -5,8 +5,6 @@ import * as express from "express";
 import * as t from "io-ts";
 import * as winston from "winston";
 
-import * as ApplicationInsights from "applicationinsights";
-
 import {
   ClientIp,
   ClientIpMiddleware
@@ -101,6 +99,7 @@ import {
   some
 } from "fp-ts/lib/Option";
 
+import { TelemetryClient } from "applicationinsights";
 import { CreatedMessageWithContent } from "../api/definitions/CreatedMessageWithContent";
 import { MessageResponseWithoutContent } from "../api/definitions/MessageResponseWithoutContent";
 import { MessageStatusValueEnum } from "../api/definitions/MessageStatusValue";
@@ -109,6 +108,17 @@ import { NotificationChannelStatusValueEnum } from "../api/definitions/Notificat
 import { TimeToLiveSeconds } from "../api/definitions/TimeToLiveSeconds";
 import { MessageStatusModel } from "../models/message_status";
 import { NotificationStatusModel } from "../models/notification_status";
+import {
+  diffInMilliseconds,
+  wrapCustomTelemetryClient
+} from "../utils/application_insights";
+
+const isProduction = process.env.NODE_ENV === "production";
+
+const getCustomTelemetryClient = wrapCustomTelemetryClient(
+  isProduction,
+  new TelemetryClient()
+);
 
 /**
  * Input and output bindings for this function
@@ -274,7 +284,7 @@ async function getMessageNotificationStatuses(
     // due to some latency, so it's better to not fail here but return an empty object
     const maybeNotification = errorOrMaybeNotification.value;
     if (isNone(maybeNotification)) {
-      winston.info(
+      winston.debug(
         `getMessageNotificationStatuses|Notification not found|messageId=${
           retrievedMessage.id
         }`
@@ -322,7 +332,6 @@ async function getMessageNotificationStatuses(
  * Returns a type safe CreateMessage handler.
  */
 export function CreateMessageHandler(
-  applicationInsightsClient: ApplicationInsights.TelemetryClient,
   messageModel: MessageModel,
   generateObjectId: ObjectIdGenerator
 ): ICreateMessageHandler {
@@ -336,6 +345,8 @@ export function CreateMessageHandler(
   ) => {
     // extract the user service
     const userService = userAttributes.service;
+
+    const startRequestTime = process.hrtime();
 
     // base appinsights event attributes for convenience (used later)
     const appInsightsEventName = "api.messages.create";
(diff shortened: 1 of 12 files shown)


lib/utils/application_insights.ts
```diff
@@ -0,0 +1,31 @@
...
+import { NonEmptyString } from "italia-ts-commons/lib/strings";
+import { ServiceId } from "../api/definitions/ServiceId";
+
+export interface ITelemetryParams {
+  readonly operationId: NonEmptyString;
+  readonly operationParentId?: NonEmptyString;
+  readonly serviceId?: ServiceId;
+}
+
+export function getApplicationInsightsTelemetryClient(
```
gunzip: ApplicationInsights.TelemetryClient() is not thread safe if used with context.keys

lib/controllers/messages.ts
```diff
@@ -337,6 +335,8 @@ export function CreateMessageHandler(
     // extract the user service
     const userService = userAttributes.service;
 
+    const startRequestTime = Date.now();
```
cloudify: You want to use a [high resolution timer](https://nodejs.org/api/process.html#process_process_hrtime_time) here - `Date.now()` may not be monotonic (i.e. affected by system clock drifts, for instance during NTP updates).

lib/created_message_queue_handler.ts
```diff
@@ -492,6 +507,17 @@ export async function index(
               newMessageWithContent.id
             }`
           );
+
+          appInsightsClient.trackEvent({
+            measurements: {
+              elapsed: Date.now() - newMessageWithContent.createdAt.getTime()
```
cloudify: this measurement is not really useful, `createdAt` gets set by `cosmos`, `Date.now()` is this system clock, the two times may not even be correlated, esp. in very short time frames (e.g. `createdAt` may be "after" `Date.now()`) - when measuring performance you need to always sample the clock from the same source

lib/utils/application_insights.ts
cloudify: I don't understand this comment, can you elaborate? 

also, if this method returns a new instance on each call, it should be called `createApplicationInsightsTelemetryClient`

lib/utils/application_insights.ts
```diff
@@ -0,0 +1,31 @@
...
+  params: ITelemetryParams,
+  commonProperties?: Record<string, string>
+): ApplicationInsights.TelemetryClient {
+  const telemetryClient = new ApplicationInsights.TelemetryClient();
+  // tslint:disable-next-line:no-object-mutation
+  telemetryClient.context.keys.operationId = params.operationId;
```
cloudify: mmh are you sure about this? from [the code](https://github.com/Microsoft/ApplicationInsights-node.js/blob/c992f39f17691f1842b225de30122856eb5acce3/Declarations/Contracts/Generated/ContextTagKeys.ts) it looks like this is the value of the `key` associated to the `operationId`, not the actual `operationId` value.

From [this example code](https://docs.microsoft.com/en-us/azure/application-insights/application-insights-correlation#role-name) it looks like the correct operation would be:

```ts
telemetryClient.context.tags[telemetryClient.context.keys.operationId] = params.operationId;
```

cloudify: also see [this code here](https://github.com/Microsoft/ApplicationInsights-node.js/blob/3abe097047d307441238e7e4a5afd5fa53fb4887/AutoCollection/HttpRequestParser.ts#L115)

lib/utils/application_insights.ts
```diff
@@ -8,7 +8,17 @@ export interface ITelemetryParams {
...
+ *
+ * When using context and/or commonProperties
+ * the TelemetryClient shares them between all threads
+ * (ie. function handlers runs). That's why we must obtain
+ * a new instance every time we want to set common values
+ * which are valid only for one run.
+ */
 export function getApplicationInsightsTelemetryClient(
+  isProduction: boolean,
   params: ITelemetryParams,
```
cloudify: would be nice to have handlers take an `(ITelemetryParams, Record<string, string>) => ApplicationInsights.TelemetryClient` and have this method as `boolean => (ITelemetryParams, Record<string, string>) => ApplicationInsights.TelemetryClient`

lib/created_message_queue_handler.ts
gunzip: this is not the value of the '_ts' parameter of cosmosdb, but a custom value set = Date.now() when creating a new message so it's fine to compare that with Date.now()

lib/utils/application_insights.ts
gunzip: yes you're right

lib/utils/application_insights.ts
gunzip: I don't get this, are you suggesting to curry the isProduction parameter ?

lib/utils/application_insights.ts
gunzip: I've put an explanation in the comments above the function. As the TelemetryClient can share some properties between track* calls, it is not safe to instantiate it at the module level (outside the index() functions) because you get a mutable shared state and results are messed up.

lib/created_message_queue_handler.ts
cloudify: yes, but same reasoning applies, since that value is saved in the db and read from the clock of a (possibly) different host - what I mean is that: you can do this, but keep in mind that any misalignment in the clock of the different nodes will be reflected in the difference of the timestamps - since we don't know how good is the clock sync of these hosts, we have no idea of the error we have in these measurements

lib/utils/application_insights.ts
cloudify: yes

lib/utils/application_insights.ts
cloudify: the problem may be caused by mutating the `context` object before calling `track` - it looks like [this approach](https://github.com/Microsoft/ApplicationInsights-node.js/issues/387#issuecomment-383648111) may be better and makes it possible to share a single client

lib/created_message_queue_handler.ts
gunzip: yes I agree, but are there any alternatives ? aside, this 'elapsed' value is usually (if not, always) expressed in terms of seconds (> 4sec. from load tests)

cloudify: one alternative could be to let AI set the timestamp (e.g. when it received the event)

what do we need this `elapsed` for in the context of our analysis?

gunzip: that's the time that passes between creating a message (in the web controller) and sending a notification (imho a pretty useful indicator).

what do you mean with "let AI set the timestamp" ?

cloudify: I mean, but I don't know if it's possible, di have AI set the timestamp of events with its own local clock (assuming that AI's clock will be consistent across events).

cloudify: my point about `elapsed` is that, assuming everything works as expected, `elapsed` will be < a few hundreds msecs - given that the smaller unit of measure is seconds, it looks it's pretty useless for our analysis

gunzip: 1. AI always set the timestamp for each event, but you cannot "extract" it at runtime and you cant have a graph with a diff in the UI (as metrics are always aggregated scalar values and cannot be dynamic). maybe you can query it someway in the analytics query UI (but not in the metric explorer), otherwise you have to check it manually.

2. the smaller unit of measure is milliseconds (Date.now / getTime), this is what you get:

image

which means that the avergae time to send a notification (every channel) is 14.09 seconds from the time the message is created calling the API (this measure will vary in production)

lib/utils/application_insights.ts
gunzip: the problem with that approach is that when you call 
`CorrelationContextManager.runWithContext(correlationContext, cb);`
you lose the return value of the wrapped callback.

lib/utils/application_insights.ts
cloudify: awesome!

digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#151202533](https://www.pivotaltracker.com/story/show/151202533): Come AMS voglio poter avere una stima sul tempo che intercorre tra l'accodamento e l'invio di una notifica, in modo da verificare il corretto funzionamento del sistema

<p align="right">
</p>

gunzip:
@cloudify this is ready for review

PR #243, opened 2018-05-28
[#157917582] return transient error in case the API proxy returns 5xx HTTP status

gunzip:
trigger a retry in case of HTTP 5xx errors. 

added tests

### DIFF
--- a/lib/webhook_queue_handler.ts
+++ b/lib/webhook_queue_handler.ts
@@ -198,7 +198,7 @@ export async function sendToWebhook(
           err.timeout
             ? TransientError(`Timeout calling API Proxy`)
             : // when the server returns an HTTP 5xx error
-              err.status && Math.floor(err.status / 100) === 5
+              err.status && err.status % 500 < 100
               ? TransientError(`Transient error calling API proxy: ${errorMsg}`)
               : // when the server returns some other type of HTTP error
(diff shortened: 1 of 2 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#157917582](https://www.pivotaltracker.com/story/show/157917582): Le API devono considerare un errore "transient" l'eventuale reset della connessione HTTP ottenuto contattando il webhook

<p align="right">
</p>

lib/webhook_queue_handler.ts
```diff
@@ -180,24 +180,28 @@ export async function sendToWebhook(
...
           err.timeout
-            ? TransientError(`Timeout calling API Proxy API`)
-            : // Fail with a permanent error in case any exception
-              // is thrown during the HTTP request
-              PermanentError(
-                `Permanent error calling API Proxy API: ${JSON.stringify(err)}`
-              )
+            ? TransientError(`Timeout calling API Proxy`)
+            : // when the server returns an HTTP 5xx error
+              err.status && Math.floor(err.status / 100) === 5
```
cloudify: to avoid the function call to `Math.floor`: `err.status % 500 < 100`

gunzip: done

PR #245, opened 2018-05-29
Fix: #224. Replace fiscal_code with tax_code.

ioggstream:
## What does this PR do

Conform the tax code specs to the ontologies defined in

  - https://github.com/italia/daf-ontologie-vocabolari-controllati

  - http://lg-patrimonio-pubblico.readthedocs.io/it/latest/index.html


api/definitions.yaml
```diff
@@ -133,8 +133,8 @@ CreatedMessageWithContent:
   properties:
     id:
       type: string
-    fiscal_code:
```
cloudify: these APIs have already been published in December and we already have a few external teams developing integrations - my suggestion is to delay this change for the v2 of the APIs

cloudify:
closing this for now as it is out of scope and we don't have bandwidth to support this change, will reopen after M1 gets completed, thanks @ioggstream

PR #248, opened 2018-05-31
[#157974182] fix: avoid permanent error when the user has no channels configured

gunzip:
Expected behavior: application Insights should log errors only for system errors.


digitalcitizenship:
## Affected stories

  * 🐞 [#157974182](https://www.pivotaltracker.com/story/show/157974182): Quando CIT non ha configurato nessun canale di notifica, non deve essere considerato un errore permanente.

<p align="right">
</p>

PR #251, opened 2018-06-02
[#157777587] feat: order messages by timestamp desc

gunzip:


lib/models/message.ts
```diff
@@ -260,7 +260,7 @@ export class MessageModel extends DocumentDbModel<
             value: fiscalCode
           }
         ],
-        query: `SELECT * FROM m WHERE (m.${MESSAGE_MODEL_PK_FIELD} = @fiscalCode)`
+        query: `SELECT * FROM m WHERE m.${MESSAGE_MODEL_PK_FIELD} = @fiscalCode ORDER BY m._ts DESC`
```
cloudify: we should probably order by ID, as is it equivalent in terms of time ordering, but is also guaranteed to be monotonic (no two elements have the same ID), thus the sorting is stable. Moreover, when we implement paging, we will be able to use a single index in the query (i.e. `WHERE m.id > last_id ORDER BY m.id DESC`, where `last_id` is last ID received by the app)

gunzip: to be able to oerde by id we must clone the id field, see
https://stackoverflow.com/questions/48710600/azure-cosmosdb-how-to-order-by-id
that's what I did in the last commit

cloudify: ops sorry what I really meant was to allow the client to query for new messages (not really a pagination of results)

PR #256, opened 2018-06-02
[#157928027] add a function to monitor queue length 

gunzip:
Runs every 30 seconds to store queue length into Application Insights metrics:

image

### DIFF
--- a/lib/queue_monitor.ts
+++ b/lib/queue_monitor.ts
@@ -17,6 +17,11 @@ const isProduction = process.env.NODE_ENV === "production";
 
 const appInsightsClient = new TelemetryClient();
 
+// needed otherwise AI will wait for the batching loop to end
+// see https://github.com/Microsoft/ApplicationInsights-node.js/issues/390
+// tslint:disable-next-line:no-object-mutation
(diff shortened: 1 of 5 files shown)


lib/queue_monitor.ts
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * customMetrics
+ *   | where * has  "queue.length"
+ *   | order by timestamp desc
+ */
+export function index(context: IContext): void {
+  const logLevel = isProduction ? "info" : "debug";
+  configureAzureContextTransport(context, winston, logLevel);
+
+  [
```
cloudify: would be nice to compose this list automatically, is there a way to enumerate the queues by querying the queues API?

digitalcitizenship:
## Affected stories

  * 🌟 [#157928027](https://www.pivotaltracker.com/story/show/157928027): Come AMS voglio monitorare il numero dei messaggi in coda per un certo intervallo temporale

<p align="right">
</p>

lib/queue_monitor.ts
cloudify: I see, never mind then :)

PR #262, opened 2018-06-08
[#158179629] feat: return sender_allowed with limited profile

gunzip:
we do not return the sender_allowed flag with full profile as with a full profile we got the whole array with blocked services

### DIFF
--- a/api/definitions.yaml
+++ b/api/definitions.yaml
@@ -255,7 +255,9 @@ LimitedProfile:
   properties:
     sender_allowed:
       type: boolean
-      default: true
+      description: |-
+        True in case the service that made the request can send 
+        messages to the user identified by this profile (false otherwise).
     preferred_languages:
       $ref: "#/PreferredLanguages"
   required:
--- a/api/definitions.yaml
+++ b/api/definitions.yaml
@@ -253,8 +253,13 @@ LimitedProfile:
     attributes.
   type: object
   properties:
+    sender_allowed:
+      type: boolean
+      default: true
     preferred_languages:
       $ref: "#/PreferredLanguages"
+  required:
+    - sender_allowed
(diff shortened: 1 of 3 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#158179629](https://www.pivotaltracker.com/story/show/158179629): Come CIT, quando un servizio vuole inviarmi una comunicazione, voglio che il servizio sappia che ho fatto optout dalle sue comunicazioni

<p align="right">
</p>

api/definitions.yaml
```diff
@@ -253,8 +253,13 @@ LimitedProfile:
     attributes.
   type: object
   properties:
+    sender_allowed:
+      type: boolean
+      default: true
```
cloudify: what does the default do in our logic?

cloudify: also worth adding a description since this is shared with the PAs

lib/controllers/profiles.ts
```diff
@@ -65,8 +72,30 @@ function toExtendedProfile(profile: RetrievedProfile): ExtendedProfile {
...
+  return (
+    blockedInboxOrChannels === undefined ||
+    blockedInboxOrChannels[serviceId] === undefined ||
+    !blockedInboxOrChannels[serviceId].has(BlockedInboxOrChannelEnum.INBOX)
+  );
+}
+
+function toLimitedProfile(
+  profile: RetrievedProfile,
+  senderAllowed: boolean
```
cloudify: `isSenderAllowed`

gunzip: cannot shadow function name

api/definitions.yaml
gunzip: removed default and added description

PR #270, opened 2018-06-11
[#156914616] feat: add payment metadata to message

gunzip:
... and forward payment amount + notice_number to the webhook (push notification)

we have to decide how to process this informations for emails


digitalcitizenship:
## Affected stories

  * 🌟 [#156914616](https://www.pivotaltracker.com/story/show/156914616): Il modello dati del messaggio va esteso con i metadati di pagamento

<p align="right">
</p>

api/definitions.yaml
```diff
@@ -97,8 +97,31 @@ NewMessage:
...
+    amount:
+      $ref: "#/PaymentAmount"
+    notice_number:
+      $ref: "#/PaymentNoticeNumber"
+PaymentNoticeNumber:
+  description: The field ["Numero Avviso"](https://pagopa-specifichepagamenti.readthedocs.io/it/latest/_docs/Capitolo7.html#il-numero-avviso-e-larchivio-dei-pagamenti-in-attesa)
+    of pagoPa, needed to identify the payment. Format is `<aux digit (1n)>[<application code> (2n)]<codice IUV (15|17n)>`.
+    See [pagoPa specs](https://www.agid.gov.it/sites/default/files/repository_files/specifiche_attuative_pagamenti_1_3_1_0.pdf) for more info on this field and the IUV.
+  type: string
+  minLength: 16
```
cloudify: would it make sense to have a regex for validation?

api/definitions.yaml
```diff
@@ -97,8 +97,31 @@ NewMessage:
...
+  maxLength: 20
+PaymentAmount:
+  description: Amount of payment in euro cent.
+  type: integer
+  minimum: 1
+  maximum: 999999999999
```
cloudify: even if pagopa accepts these kind of value, I would stay in the safe side of things and limit the amount we could handle (e.g. to say, 1000 euros, or `100000` cents)

api/definitions.yaml
```diff
@@ -162,6 +187,8 @@ CreatedMessageWithoutContent:
       $ref: "#/Timestamp"
     sender_service_id:
       type: string
+    payment_data:
```
cloudify: I would say that payment_data is a (structured) part of the message content, we are not going to store it unless the user enabled the inbox, so `MessageWithoutContent` will not include it.

api/definitions.yaml
```diff
@@ -97,8 +97,31 @@ NewMessage:
       $ref: "#/MessageContent"
     default_addresses:
       $ref: "#/NewMessageDefaultAddresses"
+    payment_data:
```
cloudify: I would put `payment_data` inside `MessageContent`, they have the same lifecycle

api/definitions.yaml
gunzip: moved under content

api/definitions.yaml
gunzip: set to 100.000 cents = 1k eur

api/definitions.yaml
gunzip: done (18 digits starting with either 0,1,2,3)

cloudify:
looks good! thanks

PR #275, opened 2018-06-14
Added docs for development

balanza:
The doc is based on my experience on setting up the application on Ubuntu. Please review if some information conflicts with other cases.


development.md
```diff
@@ -0,0 +1,169 @@
...
+```
+You can now execute and debug the application with `Debug -> Start Debugging`.
+
+## Using Docker
+
+Microsoft itself provides a docker image for running functions on node: https://hub.docker.com/r/microsoft/azure-functions-node8/.
+
+If you are familiare with `docker-compose`, you can get the app up&running with this 2 steps:
+
+1. Create a `docker-compose.yml` file with the following content:
```
gunzip: this is very nice, maybe we can include this docker-compose.yml in the root of the repo ? @cloudify

balanza: done

development.md
```diff
@@ -0,0 +1,169 @@
...
+```yml
+version: '3'
+services:
+  func:
+    image: microsoft/azure-functions-node8
+    volumes:
+      - .:/home/site/wwwroot
+    environment:
+      - AzureWebJobsScriptRoot=/home/site/wwwroot
+      - AzureWebJobsStorage=DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://storage:10000/devstoreaccount1;QueueEndpoint=http://storage:10001/devstoreaccount1;
```
cloudify: where does the key come from? is it safe to publish it/share it? :)

development.md
cloudify: yes it's a great idea, it would be nice to make it comprehensive by adding the storage emulator (is there a cosmos emulator tool?)

development.md
gunzip: yes that's the key hardcoded into the emulator

development.md
gunzip: [there is](https://github.com/Azure/azure-cosmos-db-emulator-docker), it's not clear if it works on linux hosts

balanza:
@cloudify It seems cosmos emulator isn't available for linux yet, so the option is to run a virtual windows machine. I'll do further investigation.

A related subject is *environment keys*: this doc should provide a full list of env variables used by the application. I didn't include them yet, as they are spread over code and I still didn't dig so much in code.

My intention is to keep improving this doc as well as I solve issues.

cloudify:
@balanza thanks so much for your contribution! you're right, there's not list of env vars that needs to be passed, the vars are mostly [these](https://github.com/teamdigitale/digital-citizenship/blob/master/infrastructure/azure.tf#L581) - I'm not sure whether it makes sense to duplicate the list in the readme (needs to be kept in sync 👎 ) or just link to the Terraform config... - cc @gunzip

gunzip:
probably the needed vars can be embedded into the local.settings.json example. 

these are the needed ones to run functions locally (the other vars are hardcoded into the emulator):

```sh
COSMOSDB_NAME=<cosmosdb database name>
CUSTOMCONNSTR_COSMOSDB_KEY=<cosmosdb key>
CUSTOMCONNSTR_COSMOSDB_URI=<cosmosdb url>
APPINSIGHTS_INSTRUMENTATIONKEY=<application insights key>
SCM_USE_FUNCPACK_BUILD=1
MESSAGE_CONTAINER_NAME="message-content"
MAILUP_USERNAME=<mailup username>
MAILUP_SECRET=<mailup secret>
MAIL_FROM_DEFAULT=no-reply@italia.it
WEBHOOK_CHANNEL_URL=<some webhook url>
```

development.md
```diff
@@ -0,0 +1,178 @@
...
+    ]
+}
+```
+You can now execute and debug the application with `Debug -> Start Debugging`.
+
+### Using Docker
+
+Microsoft itself provides a docker image for running functions on node: https://hub.docker.com/r/microsoft/azure-functions-node8/.
+
+If you are familiare with `docker-compose`, you can get the app up&running with this 2 steps:
```
gunzip: typo

development.md
gunzip: @balanza this description is changed a little bit since the docker-compose.yml is no included in the repo (with this PR)

balanza:
@gunzip I don't understand the change you asked me: 

>  this description is changed a little bit since the docker-compose.yml is no included in the repo (with this PR)

>

gunzip:
I meant that you don't have to 'create' the docker-compose.yml anymore since it will distributed with the sources

balanza:
@gunzip I reviewed your changes and I think everything is ok, things are cleaner and I didn't spot any typo.  Thanks for your work.

balanza:
Furthermore: I did some investigation about running the project on Linux dev machines. Although the app itself and the storage run smoothly, *Cosmosdb* cannot be executed. Viable options are:

1. Only use Windows as developer machine
2. Run [Azure Cosmosdb Emulator](https://docs.microsoft.com/en-us/azure/cosmos-db/local-emulator) on a Windows VM _(needs more investigation)_
3. Use a Cosmodb instance for develop

Option 1 is actually the easier, but forces everyone to work on Windows environment. 
Option 2 _may work fine_, but it's stressful for not-so-performant machines.
Option 3 would be the more portable, but requires a _public_ instance from _teamdigitale_ or a personal, credit-card-enabled Azure account.

That makes it hard to contribute to the project. I spent a lot of time struggling with os installation, double boot, virtualization... I think it's a big step for contributors.

Which option do you think is the one that fits more your development roadmap?

gunzip:
@balanza actually I'm on [1], [2] seems feasible, [3] is the simpler if you have an Azure account. We have considered to migrate cosmosdb api to mongodb or cassandra apis, but it's not in our roadmap atm. Apart from CosmosDB, do you have any other concerns ? I'm merging this PR anyway as we have more dev engaged into APIs and this guide will be useful for them : ) Thank you very much anyway ! 

As I've mentioned on slack, probably it's better to start contributing to side projects like the "developer portal" we're gonna start to develop this week. It will be used by the Public Administrations / developers that want to integrate with the APIs (and has a smoother path since it's just a NodeJS application with a react frontend : ). I'll keep you informed if you're interested.

PR #298, opened 2018-07-03
[#158496486] limit payment amount for each service

gunzip:
0 is default

 wip as it lacks some tests but logic is complete

### DIFF
--- a/api/definitions.yaml
+++ b/api/definitions.yaml
@@ -119,11 +119,17 @@ PaymentNoticeNumber:
   pattern: "^[0123][0-9]{17}$"
 PaymentAmount:
   description: |-
-    Amount of payment in euro cent. Even if pagoPA accepts until 999999999999 euro cents,
-    we choose to limit the amount up to 100k cents (1k euros).
+    Amount of payment in euro cent. PagoPA accepts until 999999999 euro cents.
   type: integer
   minimum: 1
-  maximum: 100000
+  maximum: 9999999999
+MaxPaymentAmount:
+  description: |-
+    Maximum amount in euro cents that a service is allowed to charge to a user.
+  type: integer
+  minimum: 0
+  maximum: 9999999999
+  default: 0
 MessageStatusValue:
   type: string
   description: |-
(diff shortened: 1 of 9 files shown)


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#158496486](https://www.pivotaltracker.com/story/show/158496486): Come AMS voglio poter impostare un limite (anche nullo) per l'importo massimo degli avvisi di pagamento per ogni singolo SAS

<p align="right">
</p>

api/definitions.yaml
```diff
@@ -119,11 +119,17 @@ PaymentNoticeNumber:
   pattern: "^[0123][0-9]{17}$"
 PaymentAmount:
   description: |-
-    Amount of payment in euro cent. Even if pagoPA accepts until 999999999999 euro cents,
-    we choose to limit the amount up to 100k cents (1k euros).
+    Amount of payment in euro cent. PagoPA accepts until 999999999 euro cents.
```
cloudify: until -> up to

until is "fino a" but in terms of time, not amount

lib/controllers/messages.ts
```diff
@@ -379,6 +380,24 @@ export function CreateMessageHandler(
...
+      paymentData.amount &&
+      (paymentData.amount as number) >
+        (userService.maxAllowedPaymentAmount as number);
+
+    // check if the service wants to charge a valid amount to the user
+    if (hasExceededAmount) {
+      return ResponseErrorValidation(
+        "Error while sending payment metadata",
+        `The requested amount exceeds the maximum allowed for this service (${
+          userService.maxAllowedPaymentAmount
```
cloudify: should probably specify that is euro cents in the error

api/definitions.yaml
```diff
@@ -119,11 +119,17 @@ PaymentNoticeNumber:
...
   type: integer
   minimum: 1
-  maximum: 100000
+  maximum: 9999999999
+MaxAllowedPaymentAmount:
```
cloudify: it may make sense to have the unit of measure as part of the name to avoid any possible misunderstanding - you know, people don't always read the docs

gunzip: we can but it will be a breaking change for some other projects that refers to these types linking to the swagger specs, so in case, we must alert all the other teams

PR #301, opened 2018-07-08
[#157244655] send welcome message

gunzip:
Reacts to messages sent to the ProfileEvents queue with one catch-all handler.

Calls the API (`submitMessageForUser`) through an HTTP request using the API-Key of the service associated to the app backend to send the welcome message.

I'd like to implement [1] using `request` method from italia-ts-commons bus in order to do that I miss the ability to add custom HTTP headers as RequestHeadersKeys are actually hardcoded.

WIP as it lacks tests.

### DIFF
--- a/lib/controllers/profiles.ts
+++ b/lib/controllers/profiles.ts
@@ -51,6 +51,7 @@ import {
   clientIPAndCidrTuple as ipTuple
 } from "../utils/source_ip_check";
 
+import { IContext } from "azure-function-express";
 import { BlockedInboxOrChannelEnum } from "../api/definitions/BlockedInboxOrChannel";
 import { ServiceId } from "../api/definitions/ServiceId";
 import {
@@ -60,6 +61,25 @@ import {
   RetrievedProfile
 } from "../models/profile";
 import { ServiceModel } from "../models/service";
+import { ContextMiddleware } from "../utils/middlewares/context_middleware";
+
+export interface IProfileCreatedEvent {
+  readonly kind: "ProfileCreatedEvent";
+  readonly fiscalCode: FiscalCode;
+  readonly newProfile: ExtendedProfile;
+}
+
+export interface IProfileUpdatedEvent {
+  readonly kind: "ProfileUpdatedEvent";
+  readonly fiscalCode: FiscalCode;
+  readonly oldProfile: ExtendedProfile;
+  readonly newProfile: ExtendedProfile;
+}
+
+interface IBindings {
+  // tslint:disable-next-line:readonly-keyword
+  profileEvent?: IProfileCreatedEvent | IProfileUpdatedEvent;
+}
 
 function toExtendedProfile(profile: RetrievedProfile): ExtendedProfile {
   return {
@@ -123,6 +143,7 @@ type IGetProfileHandler = (
  * returns a Profile or a Validation or a Generic error.
  */
 type IUpsertProfileHandler = (
+  context: IContext<IBindings>,
   auth: IAzureApiAuthorization,
   clientIp: ClientIp,
   attrs: IAzureUserAttributes,
@@ -227,7 +248,6 @@ async function createNewProfileFromPayload(
   fiscalCode: FiscalCode,
   profileModelPayload: ExtendedProfile
 ): Promise<IResponseSuccessJson<ExtendedProfile> | IResponseErrorQuery> {
-  // create a new profile
   const profile: Profile = {
     blockedInboxOrChannels: profileModelPayload.blocked_inbox_or_channels,
     email: profileModelPayload.email,
@@ -259,17 +279,22 @@ async function updateExistingProfileFromPayload(
   | IResponseErrorQuery
   | IResponseErrorInternal
 > {
+  const profile: Profile = {
+    blockedInboxOrChannels: profileModelPayload.blocked_inbox_or_channels,
+    email: profileModelPayload.email,
+    fiscalCode: existingProfile.fiscalCode,
+    isInboxEnabled: profileModelPayload.is_inbox_enabled,
+    isWebhookEnabled: profileModelPayload.is_webhook_enabled,
+    preferredLanguages: profileModelPayload.preferred_languages
+  };
+
   const errorOrMaybeProfile = await profileModel.update(
     existingProfile.id,
     existingProfile.fiscalCode,
     p => {
       return {
         ...p,
-        blockedInboxOrChannels: profileModelPayload.blocked_inbox_or_channels,
-        email: profileModelPayload.email,
-        isInboxEnabled: profileModelPayload.is_inbox_enabled,
-        isWebhookEnabled: profileModelPayload.is_webhook_enabled,
-        preferredLanguages: profileModelPayload.preferred_languages
+        ...profile
       };
     }
   );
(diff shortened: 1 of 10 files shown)


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#157244655](https://www.pivotaltracker.com/story/show/157244655): Come APP quando CIT accede all'app per la prima volta, voglio mandargli un messaggio di benvenuto per presentargli le mie funzionalità 

<p align="right">
</p>

cloudify:
Thanks, why the needs for a service bus instead of a queue storage?

gunzip:
because azure queue storage has only 1-to-1 semantic. that means you have to dispatch a tailored "welcome-message" event from the profile controller (instead of generic "profile-updated / profile-created") which means more coupling. service bus instead let us use multiple consumers (subscriptions) for the same event.

cloudify:
I don't see why we can't have the same behavior we have for the "new message -> webhook" events with webhook.

I think at this point it's better for us to keep things simpler (less moving parts) since the team is small, we're better off with less things that can break or misbehave.

The option I was thinking about was to have a simple `ProfileEventsQueue`, and have the upsert handler publish messages there (i.e. "profile-updated", "profile-created") - the handler of the `ProfileEventsQueue` acts based on the events (i.e. sending a welcome message, update the "new users of the day" table, etc...). Do you see any downsides in this approach?

gunzip:
well apart from the changed semantic (from many small functions that react to granular events to one big dispatching function that react to all events of a certain kind) the drawbacks I see are:
1. the big dispatching function will run for *all* events in the queue even if there's no action associated with that event (we'll have more executions)
1. the dispatching function will be coupled with almost everything as you have to inject cosmosdb models and so on
1. the dispatching function can grow a lot

said that, I'm not particularly biased and the switch to using a storage `ProfileEventsQueue` would be quick.

cloudify:
Yes I would keep things super simple now and as much as possible similar to what we already have (imagine new devs coming on board, less things they have to learn, the better).

Regarding the granularity of the events, we have a couple of options:

1. very specific events to specific queues (pros: more semantic, less data going through, simpler handlers, no need to query cosmos inside handlers - cons: more coupling, more logic in the producer): 
  1. `send-welcome-message` to `WelcomeEventsQueue`
  1. `profile-activated` and `profile-deactivated`  to `ProfileActivationEventsQueue`
1. very specific events to a generic queue (pros/cons same as (1), slightly more complex handler):
  1. `send-welcome-message`, `profile-activated` and `profile-deactivated` to `ProfileEventsQueue`
1. generic events to a generic queue (pros: less coupling, less logic in the producer - cons: complex handler, probably need to query cosmos for further data):
  1. `profile-created`, `profile-updated`, `profile-deleted` to `ProfileEventsQueue`
 
Personally, since queues are cheap, and the system is already complex to understand (I would not mind more semantic), I would go for (1).

gunzip:
Ok. Note that the cognitive complexity is already high (I've to suppress tslint warnings about it) for controller methods and adding more logic will increase that (so 3 may be a viable choice ?)

cloudify:
yes I would be fine with (3) too, just different tradeoffs, it's good if we can keep some complexity out of the API handlers

cloudify:
you can even have a fanout step: a handler that multiplexes events from `ProfileEventsQueue` to `WelcomeEventsQueue` and `ProfileActivationEventsQueue`

gunzip:
rewritten using azure storage queue and one catch-all queue handler for all profile related events.

lib/profile_events_queue_handler.ts
```diff
@@ -0,0 +1,104 @@
...
+);
+
+type ContextWithBindings = t.TypeOf<typeof ContextWithBindings> & IContext;
+
+async function sendWelcomeMessage(
+  url: string,
+  apiKey: string,
+  newMessage: NewMessage
+): Promise<request.Response> {
+  return request("POST", url)
```
cloudify: I wonder whether there's a way to avoid this and make a kind of "internal" call to the controller, or at least to the functions endpoint to avoid proving an API key, what do you think?

lib/profile_events_queue_handler.ts
```diff
@@ -0,0 +1,104 @@
...
+import { getRequiredStringEnv } from "./utils/env";
+
+// HTTP external requests timeout in milliseconds
+const DEFAULT_REQUEST_TIMEOUT_MS = 10000;
+
+// Whether we're in a production environment
+const isProduction = process.env.NODE_ENV === "production";
+
+// Needed to call notifications API
+const adminApiUrl = getRequiredStringEnv("ADMIN_API_URL");
```
cloudify: why is called "admin" API and not "notification API" or "public API"?

lib/profile_events_queue_handler.ts
```diff
@@ -0,0 +1,104 @@
...
+  configureAzureContextTransport(context, winston, logLevel);
+
+  winston.debug(
+    "ProfileEventsQueueHandler|Received event=%s",
+    JSON.stringify(event)
+  );
+
+  const url = `${adminApiUrl}/api/v1/messages/${event.fiscalCode}`;
+
+  const hasJustEnabledInbox =
```
cloudify: I would split this expression into sub-components as this is hard to understand, e.g.:

```ts
const isInboxEnabled = event.newProfile.is_inbox_enabled === true;
const isProfileCreated = event.kind === "ProfileCreatedEvent";
const isProfileUpdated = event.kind === "ProfileUpdatedEvent" && event.oldProfile.is_inbox_enabled === false;

const hasJustEnabledInbox = isInboxEnabled && (isProfileCreated || isProfileUpdated);
```

lib/controllers/profiles.ts
```diff
@@ -283,19 +308,17 @@ async function updateExistingProfileFromPayload(
...
-
-  const publicExtendedProfile = toExtendedProfile(profile);
-
-  return ResponseSuccessJson(publicExtendedProfile);
+  return maybeProfile.foldL<
+    IResponseErrorInternal | IResponseSuccessJson<ExtendedProfile>
+  >(
+    () =>
+      // this should never happen since if the profile doesn't exist this function
+      // will never be called, but let's deal with this anyway, you never know
```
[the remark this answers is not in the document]
cloudify: even though it may be a remote chance of this happening here too

lib/profile_events_queue_handler.ts
```diff
@@ -0,0 +1,106 @@
...
+        markdown: createWelcomeMessageMarkdown(event.newProfile),
+        subject: createWelcomeMessageSubject(event.newProfile)
+      }
+    }).getOrElseL(errs => {
+      const error = `Invalid welcome message: ${readableReport(errs)}`;
+      winston.error(`ProfileEventsQueueHandler|${error}`);
+      throw new Error(error);
+    });
+
+    // TODO: schedule retries
```
cloudify: note that retries may send multiple welcome messages as we don't have an idempotent API...

gunzip: yes, this states for all sent messages :)

PR #311, opened 2018-07-13
[#159034938] fix cosmosdb iteration and add delete + readDocuments methods

gunzip:
- fixes a problem during iteration (executeNext may return empty collection even when the iterator has more results)
- adds delete
- adds readDocuments to iterate over a collection

### DIFF
--- a/lib/utils/documentdb.ts
+++ b/lib/utils/documentdb.ts
@@ -264,6 +264,65 @@ export function readDocument<T>(
 }
 
 /**
+ * Wraps executeNext inner QueryIterator method in a Promise
+ *
+ * Returns an Either:
+ *
+ *    right(some(RetrievedDocuments)): when documents exists and are correctly retrieved
+ *    right(none): in case the iterator haven't found any document
+ *    left(QueryError): in case of error querying the database
+ */
+function executeNext<T>(
+  documentIterator: DocumentDb.QueryIterator<DocumentDb.RetrievedDocument>
+): Promise<
+  Either<
+    DocumentDb.QueryError,
+    Option<ReadonlyArray<DocumentDb.RetrievedDocument & T>>
+  >
+> {
+  return new Promise(resolve => {
+    documentIterator.executeNext((error, documents, _) => {
+      if (error) {
+        resolve(
+          left<
+            DocumentDb.QueryError,
+            Option<ReadonlyArray<T & DocumentDb.RetrievedDocument>>
+          >(error)
+        );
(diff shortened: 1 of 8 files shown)


lib/utils/documentdb.ts
```diff
@@ -274,47 +333,38 @@ export function queryDocuments<T>(
   client: DocumentDb.DocumentClient,
   collectionUri: IDocumentDbCollectionUri,
   query: DocumentDb.DocumentQuery,
-  partitionKey: string
+  partitionKey?: string
```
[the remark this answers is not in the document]
cloudify: can we have two different functions? one for standard queries that accepts a `partitionKey` and one for full scan queries, without a partition key

lib/utils/documentdb.ts
```diff
@@ -263,6 +263,65 @@ export function readDocument<T>(
...
+/**
+ * Wraps executeNext inner QueryIterator method in a Promise
+ *
+ * Returns an Either:
+ *
+ *    right(some(RetrievedDocuments)): when documents exists and are correctly retrieved
+ *    right(none): in case the iterator haven't found any document
+ *    left(QueryError): in case of error querying the database
+ */
+function executeNext<T>(
```
gunzip: this is the same method as before, but since it's now used by two different methods I've extracted it to its own function

lib/utils/documentdb.ts
```diff
@@ -416,7 +458,7 @@ export function mapResultIterator<A, B>(
                   } else {
                     resolve(
                       right<DocumentDb.QueryError, Option<ReadonlyArray<B>>>(
-                        none
+                        some([])
```
gunzip: this is because executeNext() may return empty values even if the iteration has not ended

lib/utils/documentdb.ts
```diff
@@ -540,6 +579,30 @@ export function replaceDocument<T>(
   });
 }
 
+/**
+ * Deletes a document
+ *
+ * @param documentLink  the _self link to the document
+ */
+export function deleteDocument(
```
cloudify: what we need this for?

gunzip: that's for [this one](https://github.com/teamdigitale/digital-citizenship-functions/pull/312/files#diff-1a49c09fccc54d018f5de1b082ecc1acR80)

I'd pause this PR until we converge on a safe approach for #312

gunzip:
superseeded by #312

PR #334, opened 2018-08-23
[#159003198] send two welcome messages

gunzip:
little refactor to send two messages instead of one. 

text is hardcoded atm, if we want to take it from some environment variables we must use a templating system to fill placeholders (ie. profile.name). something alike util.format or https://github.com/Matt-Esch/string-template (to have nominal format string arguments instead of positional ones)

### DIFF
--- a/lib/profile_events_queue_handler.ts
+++ b/lib/profile_events_queue_handler.ts
@@ -15,14 +15,14 @@ import {
 import { configureAzureContextTransport } from "./utils/logging";
 
 import { readableReport } from "italia-ts-commons/lib/reporters";
-import { FiscalCode, NonEmptyString } from "italia-ts-commons/lib/strings";
+import { FiscalCode } from "italia-ts-commons/lib/strings";
 import { ExtendedProfile } from "./api/definitions/ExtendedProfile";
 import { NewMessage } from "./api/definitions/NewMessage";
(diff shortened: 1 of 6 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#159003198](https://www.pivotaltracker.com/story/show/159003198): Come CIT voglio ricevere due messaggi di benvenuto appena installo la app.

<p align="right">
</p>

lib/profile_events_queue_handler.ts
```diff
@@ -29,25 +39,39 @@ const isProduction = process.env.NODE_ENV === "production";
 const publicApiUrl = getRequiredStringEnv("PUBLIC_API_URL");
 const publicApiKey = getRequiredStringEnv("PUBLIC_API_KEY");
 
+type WelcomeMessages = ReadonlyArray<
+  [(p: ExtendedProfile) => string, (p: ExtendedProfile) => string]
```
cloudify: perhaps, for consistency...

```ts
type WelcomeMessages = ReadonlyArray<(p: ExtendedProfile) => MessageContent>;
```

?

gunzip: done (I've used NewMessage that sets default values as well)

PR #352, opened 2018-11-02
Fixing styling for email's inner message.

francescozaia:
This PR aims to fix wrong line-height for any heading within `contentHtml` HTML:
screenshot 2018-11-02 at 15 37 01

As `${contentHtml}` HTML comes from Markdown and it can't be styled, we had to prevent any styling coming from its containing tags.
Due to this, I removed the global `<mj-text ...>` attribute causing its rules to be applied to any `<mj-text>` content.


francescozaia:
To be reviewed by @gunzip and @matteodesanti.

PR #373, opened 2019-03-12
fix master build and tests

gunzip:
- fix uncommitted yarn.lock
- prettify email template
- fix IResponseSuccessJsonIterator return value

### DIFF
--- a/lib/utils/response.ts
+++ b/lib/utils/response.ts
@@ -1,4 +1,5 @@
 import * as DocumentDb from "documentdb";
+import * as express from "express";
 import { isLeft } from "fp-ts/lib/Either";
 import {
   HttpStatusCodeEnum,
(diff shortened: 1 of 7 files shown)


lib/utils/response.ts
```diff
@@ -13,6 +14,9 @@ import { IResultIterator, iteratorToArray } from "./documentdb";
 export interface IResponseSuccessJsonIterator<T>
   extends IResponse<"IResponseSuccessJsonIterator"> {
   readonly value: T; // needed to discriminate from other T subtypes
+  readonly apply: (
```
cloudify: omg how could it work before?

gunzip: return value is used only in tests (that's why 2 tests fail)

PR #380, opened 2019-04-15
[#163190244] Adds accepted ToS version to user profile

chiadav:
[#163190244] L'accettazione dei Terms of Service va salvata nel profilo dell'utente

### DIFF
--- a/api/definitions.yaml
+++ b/api/definitions.yaml
@@ -322,8 +322,8 @@ ExtendedProfile:
       $ref: "#/PreferredLanguages"
     is_inbox_enabled:
       $ref: "#/IsInboxEnabled"
-    tos_version:
-      $ref: "#/TosVersion"
+    accepted_tos_version:
+      $ref: "#/AcceptedTosVersion"
     is_webhook_enabled:
       $ref: "#/IsWebhookEnabled"
     version:
@@ -338,9 +338,8 @@ IsInboxEnabled:
   description: |-
     True if the recipient of a message wants to store its content for
     later retrieval.
-TosVersion:
+AcceptedTosVersion:
   type: number
-  default: 0
   description: |-
     Version of terms of service.
 IsWebhookEnabled:
--- a/api/definitions.yaml
+++ b/api/definitions.yaml
@@ -322,15 +322,14 @@ ExtendedProfile:
       $ref: "#/PreferredLanguages"
     is_inbox_enabled:
       $ref: "#/IsInboxEnabled"
-    is_tos_accepted:
-      $ref: "#/IsTosAccepted"
+    tos_version:
+      $ref: "#/TosVersion"
     is_webhook_enabled:
       $ref: "#/IsWebhookEnabled"
     version:
       type: integer
   required:
     - is_inbox_enabled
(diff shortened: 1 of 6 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#163190244](https://www.pivotaltracker.com/story/show/163190244): L'accettazione dei Terms of Service va salvata nel profilo dell'utente

<p align="right">
</p>

fpersico:
@cloudify
There is another story to add also the accepted tos version. Can't we have just an optional accepted_tos_version field that if undefined is interpreted as no tos accepted?

api/definitions.yaml
```diff
@@ -336,6 +339,11 @@ IsInboxEnabled:
   description: |-
     True if the recipient of a message wants to store its content for
     later retrieval.
+IsTosAccepted:
+  type: boolean
```
cloudify: we need to store the version (a number) of the last ToS accepted, the app will check this number against the latest ToS and if there's a newer ToS will ask the user to accept again

api/definitions.yaml
```diff
@@ -322,6 +322,8 @@ ExtendedProfile:
       $ref: "#/PreferredLanguages"
     is_inbox_enabled:
       $ref: "#/IsInboxEnabled"
+    tos_version:
```
cloudify: perhaps `accepted_tos_version` is better, to make it clear that the user has accepted it

api/definitions.yaml
```diff
@@ -336,6 +338,11 @@ IsInboxEnabled:
   description: |-
     True if the recipient of a message wants to store its content for
     later retrieval.
+TosVersion:
+  type: number
+  default: 0
```
cloudify: don't put a default, a missing value means the user has not accepted any Tos yet

api/definitions.yaml
```diff
@@ -336,6 +338,11 @@ IsInboxEnabled:
...
+  description: |-
+    Version of terms of service.
```
cloudify:
```suggestion
    Version of latest terms of service accepted by the user.
```

api/definitions.yaml
```diff
@@ -336,6 +338,11 @@ IsInboxEnabled:
   description: |-
     True if the recipient of a message wants to store its content for
     later retrieval.
+TosVersion:
+  type: number
```
cloudify:
```suggestion
  type: number
  minimum: 1
```

cloudify: set a minimum to make it a `NonNegativeNumber`

cloudify:
@chiadav I've changed the title of this PR to make it more descriptive

lib/models/profile.ts
```diff
@@ -3,6 +3,7 @@ import * as t from "io-ts";
 import { readonlySetType, tag } from "italia-ts-commons/lib/types";
 
 import * as DocumentDb from "documentdb";
+import { AcceptedTosVersion } from "../api/definitions/AcceptedTosVersion";
```
cloudify: move this import below, along with the other imports from `"../api/definitions`

cloudify:
LGTM

PR #388, opened 2019-05-08
[#165905031] getMessages should not return rejected messages 

cloudify:
This PR introduces:

1. a pending flag on the message metadata that gets used to filter out from the result of `getMessages` the messages that are being processed (or have been rejected)
1. a new `REJECTED` status for messages that have been rejected due to a non-existent profile or a blocked sender

TODO:

- [x] fix tests

### DIFF
--- a/lib/controllers/messages.ts
+++ b/lib/controllers/messages.ts
@@ -29,6 +29,7 @@ import {
   ResponseJsonIterator
 } from "../utils/response";
 
+import { readableReport } from "italia-ts-commons/lib/reporters";
 import {
   IResponseErrorForbiddenNotAuthorized,
   IResponseErrorForbiddenNotAuthorizedForDefaultAddresses,
@@ -74,7 +75,7 @@ import {
   clientIPAndCidrTuple as ipTuple
 } from "../utils/source_ip_check";
 
-import { mapResultIterator } from "../utils/documentdb";
+import { filterResultIterator, mapResultIterator } from "../utils/documentdb";
 
 import { NotificationModel } from "../models/notification";
 import { ServiceModel } from "../models/service";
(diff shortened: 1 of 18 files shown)


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Please include a Pivotal story at the beginning of the PR title (see below).
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



  Example of PR titles that include pivotal stories:

  * single story: `[#123456] my PR title`
  * multiple stories: `[#123456,#123457,#123458] my PR title`

    
<p align="right">
</p>

lib/controllers/messages.ts
```diff
@@ -474,27 +473,14 @@ export function CreateMessageHandler(
     //
     // emit created message event to the output queue
     //
-    const errorOrNewMessageWithContent = NewMessageWithContent.decode({
```
gunzip: I suspect this decode here was needed due to some default fields that the decode() function would set when empty. If you get rid of this, default values won't be set.

lib/utils/errors.ts
```diff
@@ -5,10 +5,21 @@
  */
 
 export enum ErrorTypes {
+  // a temporaty error while processing the message (operation can be retried)
```
gunzip: typo

lib/controllers/messages.ts
cloudify: I could do a `CreatedMessageEvent.decode`, it should be equivalent as it's composed of the same sub-types as  `NewMessageWithContent`, what do you think? I'm fixing the tests now, perhaps the issue will emerge there

gunzip: that's right

cloudify: just found the problem with a test not passing, I confirm that using `CreatedMessageEvent.decode` fixes the problem

